PR #12, opened 2021-10-13
chore(deps-dev): bump dependent package and tools versions

jenstroeger:


jenstroeger:
We can keep this PR open for another day or two, in case Dependabot pushes more updates.

Unfortunately, Dependabot does _not_ consider checking & updating the [pre-commit-config.yml](https://github.com/jenstroeger/python-package-template/blob/master/.pre-commit-config.yaml) file (see issue https://github.com/dependabot/dependabot-core/issues/1524). But we’d address that using a dedicated Action as described in [this blog](https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/). Separate PR, perhaps?

PR #48, opened 2021-11-22
chore: tweak project settings

jenstroeger:
As we’re using the repository for various projects, we continue to find details that need tweaking… so here’s the first batch of ’em 🤓

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,7 @@ repos:
       - id: mypy
         files: ^src/package/|^tests/|setup.py
         additional_dependencies: [types-setuptools==57.4.2,hypothesis==6.30.1]
-        args: [--config-file, mypy.ini]
+        args: [--ignore-missing-imports, --config-file, mypy.ini]
 
   # Enable a whole bunch of useful helper hooks, too.
   # See https://pre-commit.com/hooks.html for more hooks.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,7 @@ repos:
       - id: mypy
         files: ^src/package/|^tests/|setup.py
         additional_dependencies: [types-setuptools==57.4.2,hypothesis==6.30.1]
-        args: [--ignore-missing-imports, --config-file, mypy.ini]
+        args: [--config-file, mypy.ini]
 
   # Enable a whole bunch of useful helper hooks, too.
   # See https://pre-commit.com/hooks.html for more hooks.
--- a/setup.py
+++ b/setup.py
@@ -27,11 +27,11 @@ setuptools.setup(
     description="",
     long_description=README,
     long_description_content_type="text/markdown",
-    url="",
-    author="",
-    author_email="",
-    # maintainer=
-    # maintainer_email=
+    url="https://project.url/",
+    author="Author",
+    author_email="author@email",
+    maintainer="Maintainer",
+    maintainer_email="maintainer@email",
     license=LICENSE,
     # https://pypi.org/classifiers/
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -36,17 +36,24 @@ setuptools.setup(
     # https://pypi.org/classifiers/
     classifiers=[
         "Development Status :: 1 - Planning",
+        "Intended Audience :: Developers",
+        "Natural Language :: English",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3 :: Only",
+        "Programming Language :: Python :: Implementation :: CPython",
     ],
+    python_requires=">=3.9",
     keywords="",
     project_urls={
         "Homepage": "https://foo.bar/",
     },
-    package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
-    python_requires=">=3.9",
+    package_dir={"": "src"},
+    package_data={
+        "package": ["py.typed"],
+    },
     include_package_data=True,
     install_requires=[],
     extras_require={
@@ -64,9 +71,6 @@ setuptools.setup(
             "tox==3.24.4",
         ],
         "docs": ["sphinx==4.3.1"],
-    },
-    package_data={
-        "package": ["py.typed"],
     },
     options={},
     platforms="",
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -27,11 +27,11 @@
     description="",
     long_description=README,
     long_description_content_type="text/markdown",
-    url="",
-    author="",
-    author_email="",
-    # maintainer=
-    # maintainer_email=
+    url="https://project.url/",
```
acrobat888, line 30: You might want to update the README file now that these are no longer "" (empty) to call out to the engineers to change them after "forking" or making a new repo.

acrobat888 (approved):
Made a comment but not critical

setup.py
jenstroeger: Oh yes, good point!

.pre-commit-config.yaml
```diff
@@ -15,27 +15,27 @@ repos:
 
   # Sort imports.
   - repo: https://github.com/pycqa/isort
-    rev: 5.9.3
+    rev: 5.10.1
```
behnazh, line 18: Are you planning to manually update `pre-commit` hooks?

.pre-commit-config.yaml
```diff
@@ -66,8 +66,8 @@ repos:
     hooks:
       - id: mypy
         files: ^package/|^tests/|setup.py
-        additional_dependencies: [types-setuptools==57.4.2,hypothesis==6.24.2]
-        args: [--config-file, mypy.ini]
+        additional_dependencies: [types-setuptools==57.4.2,hypothesis==6.27.0]
+        args: [--ignore-missing-imports, --config-file, mypy.ini]
```
behnazh, line 70: This config is now diverging from the [main mypy.ini](https://github.com/jenstroeger/python-package-template/blob/updates-and-tweaks/mypy.ini#L25) config?

.pre-commit-config.yaml
jenstroeger: Well, there’s [autoupdate](https://pre-commit.com/#updating-hooks-automatically) but I guess we’ll have to run that every so often…

.pre-commit-config.yaml
jenstroeger: Hm, well, according to [this](https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit):

> By default, mypy will run with `mypy --ignore-missing-imports`, `pre-commit` runs `mypy` from an isolated virtualenv so it won't have access to those. To change the arguments, override the `args` as follows:

this changes ensures both are equal, no?

.pre-commit-config.yaml
behnazh: I recommend to add it as an `Actions workflow` and schedule it to run and create a PR every month for example.

.pre-commit-config.yaml
behnazh: Hm, according to [mypy's documentation](https://mypy.readthedocs.io/en/latest/config_file.html#confval-ignore_missing_imports), the default value for `ignore_missing_imports` is `False` though.

.pre-commit-config.yaml
jenstroeger: As in… a [scheduled job](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onschedule)?

.pre-commit-config.yaml
jenstroeger: Then we’ll have to try it out to find it out 🤓

.pre-commit-config.yaml
behnazh: > 

Yes, that would be great I think.

jenstroeger: @behnazh, should we open an issue for this?

jenstroeger: Something like the [pre-commit-autoupdate](https://github.com/marketplace/actions/pre-commit-autoupdate) action? Looks like heaps of options are available here.

setup.py
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/48/commits/680f985144638c12c37ddeca08cba9be13c52375.

.pre-commit-config.yaml
jenstroeger: Considering that with this change we also add
```ini
# Ignore the following imports because they have no types available.
[mypy-pytest]
ignore_missing_imports=True
```
to the `mypy.ini` file I’m tempted to remove `--ignore-missing-imports` command-line option. Just feels like duplicating settings.

.pre-commit-config.yaml
behnazh: Yes, opening a new issue sounds good. I prefer not using a third-party action for the pre-commit autoupdate though because it's fairly simple.

.pre-commit-config.yaml
behnazh: > to the mypy.ini file I’m tempted to remove --ignore-missing-imports command-line option. Just feels like duplicating settings.

I wonder if the `--ignore-missing-imports` in the [pre-commit-config.yaml](https://github.com/jenstroeger/python-package-template/blob/updates-and-tweaks/.pre-commit-config.yaml#L70) takes precedence and also applies to [mypy-src](https://github.com/jenstroeger/python-package-template/blob/updates-and-tweaks/mypy.ini#L25)?

.pre-commit-config.yaml
jenstroeger: Issue https://github.com/jenstroeger/python-package-template/issues/63.

.pre-commit-config.yaml
jenstroeger: From [the docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-config-file):

> By default settings are read from `mypy.ini`, `.mypy.ini`, `pyproject.toml`, or `setup.cfg` in the current directory. Settings override mypy’s built-in defaults and command line flags can override settings.

behnazh: That's true. And that's why I suspect the pre-commit hook would apply  `ignore-missing-imports` to not only the tests by the main src modules too. We can try removing the cmd option and see if the pre-commit hook still goes through.

jenstroeger: I agreeably agree.

behnazh (approved):

PR #64, opened 2021-12-07
ci: remove unnecessary semantic-release tool from dev environment

### LINKED ISSUE #62: Fix semantic-release tool inconsistency.
We have specified this dependency:

https://github.com/jenstroeger/python-package-template/blob/b36c1d0e3cd288635fc16442f5c848b4abff3a9a/setup.py#L63

whereas the Github action uses this one:

https://github.com/jenstroeger/python-package-template/blob/b36c1d0e3cd288635fc16442f5c848b4abff3a9a/.github/workflows/check-and-release-main.yml#L44

So, the locally installed [semantic-release tools](https://github.com/relekang/python-semantic-release) may not match the ones used by the Action when running a CI workflow.

I think there are several ways of addressing this:

1. Pin the Github Action to the same version as specified by `setup.py`. That’ll introduce duplication, though, and feels tedious.
2. Remove the dependency from `setup.py` because the tools aren’t used often locally; if they are, a manual installation may suffice.
3. Tools like [act](https://github.com/nektos/act) may be a good replacement for 2 while ensuring a single point of specifying that dependency.

Thoughts?

jenstroeger:
This PR closes #62 but I wanted to discuss some more. Specifying

https://github.com/jenstroeger/python-package-template/blob/de5c51b08c6bd97880cc24594a928e19c5932ecb/.github/workflows/check-and-release-main.yml#L44

doesn’t actually _pin_ a version and instead every run of this Action uses whatever changes are at the head of the `master` branch. That, I think, violates some of the [SLSA build requirements](https://slsa.dev/requirements#build-requirements) like “Reproducible”.

📌 it?

### DIFF
--- a/.github/workflows/check-and-release-main.yml
+++ b/.github/workflows/check-and-release-main.yml
@@ -41,7 +41,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Python Semantic Release
-      uses: relekang/python-semantic-release@master
+      uses: relekang/python-semantic-release@v7.23.0
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         pypi_token: ""  # We don't publish to PyPi: ${{ secrets.PYPI_TOKEN }}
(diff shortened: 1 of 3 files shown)


behnazh:
> pushpin it?

Yes please. I think all the Actions dependencies should be pinned for reproducibility.

jenstroeger:
> Yes please. I think all the Actions dependencies should be pinned for reproducibility.

Commit https://github.com/jenstroeger/python-package-template/pull/64/commits/d93db78b042bd38430880a156a346d8d0023e194.

behnazh (approved):

Issue #1, opened 2021-09-29
Switch from pip to pipenv or Poetry.


jenstroeger:
Currently the package uses [pip](https://github.com/pypa/pip) and Python’s [setuptools](https://github.com/pypa/setuptools) as described in [Packaging and distributing projects](https://packaging.python.org/guides/distributing-packages-using-setuptools/). However, problems with recursive dependency checks, deep version collisions, or cumbersome pinning make working with pip somewhat unreliable.

Consider switching to [pipenv](https://github.com/pypa/pipenv) or [Poetry](https://github.com/python-poetry/poetry) instead…

jenstroeger:
Other packaging and env-management tools recommended during conversations with @hynek and @behnazh, are [flit](https://github.com/takluyver/flit) and [asdf](https://github.com/asdf-vm/asdf) and [pyenv](https://github.com/pyenv/pyenv).

jenstroeger:
Consider also the discussion on [PEP 665: A file format to list Python dependencies for reproducibility of an application](https://discuss.python.org/t/pep-665-take-2-a-file-format-to-list-python-dependencies-for-reproducibility-of-an-application/11736).

behnazh:
The PEP 665 makes a lot of sense. It also mentions how [npm](https://www.npmjs.com/) has solved the reproducibility problem by using [package.json](https://docs.npmjs.com/cli/v6/configuring-npm/package-json).

But for the time being Poetry seems to be doing well enough in dependency management and reproducibility I think.

jenstroeger:
Interesting conversation [regarding `flit` on python.org](https://discuss.python.org/t/should-flit-become-a-pypa-project/11854).

behnazh:
> Interesting conversation [regarding `flit` on python.org](https://discuss.python.org/t/should-flit-become-a-pypa-project/11854).

Interesting indeed. I didn't find any solid reasoning in the thread for Flit though. In particular, I tend to disagree with [this argument](https://discuss.python.org/t/should-flit-become-a-pypa-project/11854/5):

> my default is that any packaging-related project that wants to be part of PyPA should be welcomed. We shouldn’t need to justify including projects, rather we should require justification from people who want to say “no” to a request to join.

Welcoming any project without requiring justifications brings us to where we are now, which might be convenient for some developers but quite chaotic compared to [npm](https://www.npmjs.com/).

jenstroeger:
[...]

PR #74, opened 2021-12-14
ci: add Makefile to enable the “Scripted Build” rule for SLSA Level 1

### LINKED ISSUE #69: Moving towards SLSA Level 1.
@behnazh, if I understand you correctly in related issue https://github.com/jenstroeger/python-package-template/issues/25 then we need a [scripted build](https://slsa.dev/requirements#scripted-build) of sorts reach [SLSA Level 1](https://slsa.dev/levels#level-requirements) for this repo. For Python packages (What about applications?) that’s described on the [Creating Built Distributions](https://docs.python.org/3/distutils/builtdist.html) — is that what you mean?

It sounds like the script itself is required, but not the built artifact, e.g.
```bash
#!/usr/bin/env bash

python setup.py bdist
```
which would create e.g. a `dist/package-1.1.2.macosx-10.14-x86_64.tar.gz` artifact; or using the [`sdist`](https://docs.python.org/3/distutils/sourcedist.html) to create a `dist/package-1.1.2.tar.gz` artifact. (See also the [setuptools](https://setuptools.pypa.io/en/latest/userguide/index.html) documentation.)

jenstroeger:
This PR closes #69. For now I leave it in Draft mode because we probably want to update the [README](https://github.com/jenstroeger/python-package-template/blob/main/README.md) and @behnazh really really wants to add the appropriate Action.


Makefile
```diff
@@ -0,0 +1,21 @@
...
+dist:	dist/.dist
+dist/.dist:
+		python setup.py bdist
+		python setup.py sdist
+		touch dist/.dist
+
+.PHONY:	check
+check:
+		pylint src/ tests/
+		mypy src/ tests/
```
jenstroeger, line 16: Alternatively, we could consider calling `pre-commit run --all-files` here.

jenstroeger:
Also, hmm, what about [Lucas-C/pre-commit-hooks-go](https://github.com/Lucas-C/pre-commit-hooks-go) 🤔

Makefile
```diff
@@ -0,0 +1,21 @@
+#! /usr/bin/env make -f
+# This Makefile assumes an activated Python virtual environment.
```
behnazh, line 2: So, the assumption is that the GitHub Actions workflow is in charge of setting up the virtual environment and installing the dependencies, right? And a local build may or may not be producing the same artifact?

Makefile
```diff
@@ -0,0 +1,21 @@
...
+.PHONY:	all
+all:	dist
+
+dist:	dist/.dist
+dist/.dist:
+		python setup.py bdist
+		python setup.py sdist
+		touch dist/.dist
+
+.PHONY:	check
```
behnazh, line 13: Can we make this check a requirement for the `dist` step?

Makefile
behnazh: How about `pytest`? And again we need to decide between `pre-commit` and the CI tests?

Makefile
jenstroeger: > So, the assumption is that the GitHub Actions workflow is in charge of setting up the virtual environment and installing the dependencies, right?

Well, yes:

https://github.com/jenstroeger/python-package-template/blob/7292ee9804ba6cae77b63726ae9691e9aeb154e4/.github/workflows/pull-request.yml#L34-L43

> And a local build may or may not be producing the same artifact?

Unfortunately, yes, I think. The user of the local venv needs to make sure that the venvs and the venv setup are in sync: after modifying `setup.py` to update dependent package versions I need to [`pip install --upgrade`](https://pip.pypa.io/en/stable/user_guide/#only-if-needed-recursive-upgrade) and  [`tox --recreate`](https://tox.wiki/en/latest/config.html#cmdoption-tox-r) to pick up the new package.

Makefile
jenstroeger: We can, but that’s a little annoying: `check` doesn’t write a persistent output which mean it _always_ runs. If I add it as a dependency to `dist`
```Makefile
dist: dist/.dist
dist/.dist: check
	python setup.py bdist
	python setup.py sdist
	touch dist/.dist
```
then checks run whether a distribution artifact exists or not. Touching a phony file like we do with `dist/.dist` seems to add more clutter than is worth.

jenstroeger: Note that the Stripe folks also use a [Makefile](https://github.com/stripe/stripe-python/blob/master/Makefile)…

behnazh: > We can, but that’s a little annoying: `check` doesn’t write a persistent output which mean it _always_ runs. If I add it as a dependency to `dist`

I think we need to have different options for building and packaging. For the `build` , we could pass a `skipTests` flag to optionally turn it off when needed.

Also, I think we can use this `Makefile` to perform the checks and tests in various Actions workflows, instead of doing the checks as separate workflow jobs. That would let us avoid copying these checks, which can be error-prone and problematic :

https://github.com/jenstroeger/python-package-template/blob/7292ee9804ba6cae77b63726ae9691e9aeb154e4/.github/workflows/pull-request.yml#L29-L53

https://github.com/jenstroeger/python-package-template/blob/7292ee9804ba6cae77b63726ae9691e9aeb154e4/.github/workflows/check-and-release-main.yml#L10-L34

jenstroeger:
@behnazh please take a look at the updated Makefile. Most of the above feedback went into it, and then some.

I think we ought to split the `make clean` into two — one `clean` which removes only the generated dist and docs files, and one `deep-clean` (or “nuke”).

Makefile
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/139ef23b1619e8b00394129043c3c621018f54ed.

Makefile
jenstroeger: Please see commit https://github.com/jenstroeger/python-package-template/pull/74/commits/139ef23b1619e8b00394129043c3c621018f54ed, and my [comment below](https://github.com/jenstroeger/python-package-template/pull/74#issuecomment-1031115562).

Makefile
```diff
@@ -0,0 +1,40 @@
...
+# This Makefile assumes an activated Python virtual environment.
+
+# Make sure that Python's virtual environment is activated, and
+# all tools are available.
+ifndef VIRTUAL_ENV
+  $(error Please activate the Python virtual environment)
+endif
+
+.PHONY:	all
+all:	dist
```
behnazh, line 11: I'm not sure if only running `dist` here makes sense. When I run `make all` I expect to run all the stages.

We could change `dist` to `build` and then add all the following stages to `all`?

Makefile
```diff
@@ -0,0 +1,40 @@
...
+.PHONY: docs
+docs:	docs/_build/html/index.html
+docs/_build/html/index.html:
+		cd docs && make html && cd ..
+
+.PHONY:	clean
+clean:
+		rm -fr .hypothesis .coverage .mypy_cache .pytest_cache
+		rm -fr dist/*.tar.gz dist/.dist docs/_build
+		find src/ -name "__pycache__" -exec -fr {} +
```
behnazh, line 39: I think you forgot to add `rm`?

```
find src/ -name "__pycache__" -exec rm -fr {} +
```

behnazh:
> I think we ought to split the `make clean` into two — one `clean` which removes only the generated dist and docs files, and one `deep-clean` (or “nuke”).

I think that's a good idea. Especially, the `pre-commit clean` should be in the `deep-clean`, otherwise it's an overkill.

behnazh:
In general I agree with your suggestions, and adding `make install`, `make setup` and `make deep-clean` would be great.

Makefile
jenstroeger: At least GNU Make defines a set of [Standard Targets](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html)<sup>*</sup> and it defines the “dummy target” `all` as

> Compile the entire program. This should be the default target. This target need not rebuild any documentation files; […]

Considering that this is a Python package where we don’t actually compile anything, I used the `dist` target as the closest sensible target for the `all` dummy:

> Create a distribution tar file for this program. The tar file should be set up so that the file names in the tar file start with a subdirectory name which is the name of the package it is a distribution for. This name can include the version number.

> @behnazh When I run `make all` I expect to run all the stages.

With “stages” do you mean build targets? If the package has been built already, then what would you expect `make all` to do?

—————
<sup>*</sup> See also [this Stackoverflow question](https://stackoverflow.com/questions/2514903/what-does-all-stand-for-in-a-makefile).

Makefile
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/9dd92bcfc8c67670910356713b5382f7d34a2fc1.

Makefile
behnazh: As you mentioned because Python packages don't compile anything, it might not make sense to strictly follow the conventions in GNU Make.

I expect `make all` to run the `quick-check`, `check`, `test`, and if the package has not already been built `dist` (`build`) targets.

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/68d6a919002d5c9f327dc19f57fbd6a97bbae32c.

behnazh (commented):
LGTM. Can you please document that we can now use `make` in  `README.md`? It might be the right time to have a contribution section too, but that can be done in a separate PR.

behnazh:
I tried to use the `Makefile` in an action and observed a few problems:

- We need to activate virtual env per rule
- It would be good to add the `pip install` commands as a `make install` rule
- The Actions workflows should be updated in this PR to make sure the `Makefile` runs correctly. I have noticed inconsistencies in running `make check` locally and in the Actions workflow.

jenstroeger:
I still need to document the Makefile use in our README, will do so next.

> * We need to activate virtual env per rule

Done, commit https://github.com/jenstroeger/python-package-template/pull/74/commits/5bc3d50c4cac554a3e7e7cf6171ab93f295d6dd3.

> * It would be good to add the `pip install` commands as a `make install` rule

Usually `make install` installs the built artifacts on the host which isn’t what you intend here, I think. Instead I added a `setup` goal so that `make setup` creates a virtual environment and installs all dependent packages into it.

> * The Actions workflows should be updated in this PR to make sure the `Makefile` runs correctly. I have noticed inconsistencies in running `make check` locally and in the Actions workflow.

I agree, and I’ll take a look.

behnazh:
That's wonderful. It's so smooth and easy now :smiley:

behnazh:
> Should we have dedicated make setup and make setup-test and make setup-dev and make setup-docs and make setup-all goals?

I don't think so. That seems an overkill to me.

behnazh:
It might be a good idea to remove `.venv/` as part of `make nuke` or a variation of it to have a fresh virtual environment.

Makefile
```diff
@@ -0,0 +1,67 @@
...
+test:
+	. .venv/bin/activate && \
+	pre-commit run pytest --hook-stage push
+
+.PHONY: docs
+docs: docs/_build/html/index.html
+docs/_build/html/index.html:
+	$(MAKE) -C docs/ html
+
+.PHONY: dist-clean clean
```
behnazh, line 56: I don't think this target requires an activated virtual environment. Especially when things go wrong, we should allow a cleanup nonetheless.

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/90b0a842b050937f9446f96e276e5c854d9e9647.

jenstroeger:
Oh… we need to adjust the Actions to use the Makefile!

Makefile
```diff
@@ -0,0 +1,93 @@
...
+	pre-commit run pylint --all-files && \
+	pre-commit run mypy --all-files
+check:
+	. .venv/bin/activate && \
+	pre-commit run --all-files
+
+.PHONY: test
+test:
+	. .venv/bin/activate && \
+	pre-commit run pytest --hook-stage push
```
jenstroeger, line 59: We _should_ run
```
python -W always::DeprecationWarning -m pytest
```
to make sure warnings are issued, see also the [`-W` command-line option](https://docs.python.org/3/using/cmdline.html#cmdoption-W).

jenstroeger, line 59: We _should_ run
```
python -W always::DeprecationWarning -m pytest
```
to make sure warnings are issued, see also the [`-W` command-line option](https://docs.python.org/3/using/cmdline.html#cmdoption-W).

Makefile
```diff
@@ -0,0 +1,93 @@
...
+	rm -fr .venv/build/* .venv/dist/*
+clean: dist-clean
+	rm -fr .hypothesis .coverage .mypy_cache .pytest_cache
+	rm -fr docs/_build
+
+.PHONY: nuke-caches nuke
+nuke-caches: clean
+	find src/ -name __pycache__ -exec rm -fr {} +
+	find tests/ -name __pycache__ -exec rm -fr {} +
+	if [[ -f .venv/bin/pre-commit ]]; then . .venv/bin/activate && pre-commit clean; fi
```
jenstroeger, line 89: Hmm, perhaps run `pre-commit` directly without activating:
```bash
if [[ -f .venv/bin/pre-commit ]]; then .venv/bin/pre-commit clean; fi
```

jenstroeger, line 89: Hmm, perhaps run `pre-commit` directly without activating:
```bash
if [[ -f .venv/bin/pre-commit ]]; then .venv/bin/pre-commit clean; fi
```

Makefile
behnazh: Running `pytest` with `-W`is turned off [in the `.pre-commit-config.yaml` too](https://github.com/jenstroeger/python-package-template/blob/main/.pre-commit-config.yaml#L151). Do you mean we should turn it on in both places and in the CI?

UPSTREAM_README.md
```diff
@@ -94,15 +86,24 @@ If you’d like to start your own Python project from scratch, you can either co
...
+```
+
+And finally, activate your virtual environment (the period here is a [shorthand for the `source` command](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/html_node/Bourne-Shell-Builtins.html#index-_002e)):
+
 ```bash
-python3.10 -m venv .
-source ./bin/activate
-pip install --upgrade pip
-pip install --editable .[dev,test,docs]
+. .venv/bin/activate
```
behnazh, line 103: Doesn't `make setup` already activate the `venv`?

jenstroeger: Every recipe in a Makefile runs in its own subprocess, and so when the Makefile is done the process is done and with it the activated environment for that process.

This line activates the venv for the shell process itself.

Makefile
jenstroeger: I haven’t figured out how to run that in pre-commit 😕

Makefile
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/48b6114b21243fd92ae7913436a2f478b214990c.

jenstroeger:
From the [comment above](https://github.com/jenstroeger/python-package-template/pull/74#issuecomment-1066714243):

https://github.com/jenstroeger/python-package-template/blob/e9fabbe583c3f849f9e04a0658171761b5112ec7/.github/workflows/pull-request.yaml#L42-L47

It looks like that change [fails on Windows](https://github.com/jenstroeger/python-package-template/runs/5611261144) and that’s because the `bin/activate` apparently is `Scripts/activate` ([docs](https://docs.python.org/3/using/windows.html#finding-the-python-executable)). Really? 😳

jenstroeger:
Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/34b8953e24f1a000098e0377fbb276dc8f829bcd adds two new targets:

- `upgrade` which reinstalls the package and its dependencies based on the [setup.py](https://github.com/jenstroeger/python-package-template/blob/main/setup.py) file; and
- `requirements` which re/builds a [requirements.txt](https://pip.pypa.io/en/latest/reference/requirements-file-format/) file including integrity hashes (see also issue https://github.com/peterbe/hashin/issues/139), as well as the [`pip --hash` option](https://pip.pypa.io/en/stable/cli/pip_hash/) and the [Hash-checking mode](https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode)).

The latter is useful (and necessary for SLSA) when used in combination with
```bash
pip install --require-hashes -r requirements.txt
```

Makefile
```diff
@@ -0,0 +1,109 @@
+#! /usr/bin/env make -f
```
behnazh, line 1: Could you please add documentation to the `Makefile`?

behnazh:
> From the [comment above](https://github.com/jenstroeger/python-package-template/pull/74#issuecomment-1066714243):
> 
> https://github.com/jenstroeger/python-package-template/blob/e9fabbe583c3f849f9e04a0658171761b5112ec7/.github/workflows/pull-request.yaml#L42-L47
> 
> It looks like that change [fails on Windows](https://github.com/jenstroeger/python-package-template/runs/5611261144) and that’s because the `bin/activate` apparently is `Scripts/activate` ([docs](https://docs.python.org/3/using/windows.html#finding-the-python-executable)). Really? flushed

Well, we might need to decide if activating the virtual environment in the `Makefile` is a good idea. It might be better to leave that to the user?

Makefile
jenstroeger: Do we want to add comments into the Makefile, or document it in more detail in our README?

jenstroeger:
> > It looks like that change [fails on Windows](https://github.com/jenstroeger/python-package-template/runs/5611261144) […]
>
> Well, we might need to decide if activating the virtual environment in the `Makefile` is a good idea. It might be better to leave that to the user?

Hmm, good point. I’m tempted to leave the environment activation to the user because:

- a user can then use multiple venvs for the same repo and switch between them; and
- the Makefile would work on whichever environment the user has activated, instead of a presumed `.venv` default; and
- it would unblock Windows again.

Makefile
behnazh: > Do we want to add comments into the Makefile, or document it in more detail in our README?

I think both?

Makefile
```diff
@@ -0,0 +1,109 @@
...
+
+.PHONY: dist-clean clean
+dist-clean:
+	rm -fr .venv/build/* .venv/dist/*
+clean: dist-clean
+	rm -fr .hypothesis .coverage .mypy_cache .pytest_cache
+	rm -fr docs/_build
+
+.PHONY: nuke-caches nuke
+nuke-caches: clean
```
behnazh, line 102: I think we need to remove `.venv/` completely to allow a fresh setup.

behnazh, line 102: `make nuke-caches` still removes `pre-commit` caches for all my projects :-(

jenstroeger: Yes I noticed the same, but…

https://github.com/jenstroeger/python-package-template/blob/34b8953e24f1a000098e0377fbb276dc8f829bcd/Makefile#L101-L109

Looking at [the code](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/commands/clean.py), I think the [`clean` option ](https://pre-commit.com/#pre-commit-clean) actually cleans out the _entire_ “store” which impacts all repositories that use pre-commit. Sigh.

Also, I think we should remove the `pre-commit uninstall` because it’s managed by the package setup as part of the `hooks` extra:

https://github.com/jenstroeger/python-package-template/blob/34b8953e24f1a000098e0377fbb276dc8f829bcd/setup.py#L60

Hmm, would be nice if pre-commit would be able to manage/clean caches for a particular repo only 🤔

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/c6dd9925855fffa5cfd257970224a74e41c19481.

jenstroeger:
> Hmm, good point. I’m tempted to leave the environment activation to the user because:

Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/c3aa54475bca497ff0ca2d0e6b05be9783620e6f.

Makefile
jenstroeger: Commits https://github.com/jenstroeger/python-package-template/pull/74/commits/aa69524feffc597dc269fbb1d24c31902e7856f1 and https://github.com/jenstroeger/python-package-template/pull/74/commits/579f44cc3716fadd850d2ba10f9b12156275a647.

jenstroeger:
Regarding the “build a distribution artifact” code:

https://github.com/jenstroeger/python-package-template/blob/579f44cc3716fadd850d2ba10f9b12156275a647/Makefile#L84-L92

I think we should:
- store the artifacts in `dist/` instead of `$(VIRTUAL_ENV)/dist/` to keep access to the artifacts more intuitive; and
- make sure that the artifact name includes the actual Python version number instead of just `py3`.

I also want to add a section “Package vs. Application” to the README based on https://github.com/jenstroeger/python-package-template/issues/1#issuecomment-1075684310 which also mentions the integrity hashes generated by the Makefile, and how to use them

Makefile
jenstroeger: Hmm, looks like this works:
```yaml
    name: Run unit tests
    entry: bash -c 'python -W always::DeprecationWarning -m pytest'
    language: system
```

jenstroeger: Even simpler: commit https://github.com/jenstroeger/python-package-template/pull/74/commits/e2fd36fb6e5027b1c550ebef278a147318a606dc.

Makefile
```diff
@@ -0,0 +1,116 @@
...
+	if [ ! -z "${VIRTUAL_ENV}" ]; then \
+	  echo "Found an activated Python virtual environment, exiting" && exit 1; \
+	fi
+	if [ -z "${PYTHON}" ]; then \
+	  echo "Creating virtual envirnoment in .venv/ for python3.10" \
+	  python3.10 -m venv --upgrade-deps .venv; \
+	else \
+	  echo "Creating virtual envirnoment in .venv/ for ${PYTHON}" \
+	  ${PYTHON} -m venv --upgrade-deps .venv; \
+	fi
```
behnazh, line 41: You need to add  `&&` to run the commands proceeding `echo`. At the moment `make venv` just echoes strings and does nothing:

```
	if [ -z "${PYTHON}" ]; then \
	  echo "Creating virtual envirnoment in .venv/ for python3.10" \
	  && python3.10 -m venv --upgrade-deps .venv; \
	else \
	  echo "Creating virtual envirnoment in .venv/ for ${PYTHON}" \
	  && ${PYTHON} -m venv --upgrade-deps .venv; \
	fi
```

behnazh (changes requested):
Very nice. Just one bug! ;-)

.github/workflows/check-and-release-main.yaml
```diff
@@ -15,23 +15,19 @@ jobs:
...
-    - name: Run pytest
-      run: python -W always::DeprecationWarning -m pytest
+      run: make setup
+      env:
+        PYTHON: python
+    - name: Run checks
+      run: make check
+    - name: Run tests
+      run: make test
   release:
```
behnazh, line 31: Looks like you've forgotten to upgrade `actions/checkout` to `v3` in the `release` job?

Makefile
jenstroeger: Instead of `&&` I separated the two commands with a `;` because the echo can’t (shouldn’t ever) fail so there’s no need to make the second command conditional on the first one ([docs](https://www.gnu.org/software/bash/manual/bash.html#Lists-of-Commands)).

Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/0ae031343262a69a0d0b65be7b5d8c44d9ced26d.

.github/workflows/check-and-release-main.yaml
jenstroeger: Nice catch! Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/7f28089faca3156b29661134b86083b735e6c4d2.

jenstroeger:
@behnazh, I wonder if we should create the following dependencies for some of the targets:
```diff
diff --git a/Makefile b/Makefile
index 65ed37e..9081c35 100644
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ setup: upgrade
 # Install or upgrade an existing virtual environment based on the
 # package dependencies declared in setup.py.
 .PHONY: upgrade
-upgrade:
+upgrade: setup.py
        pip install --upgrade pip
        pip install --editable .[hooks,dev,test,docs]
 
@@ -60,7 +60,7 @@ upgrade:
 # hashes for all packages currently installed in the virtual environment.
 .PHONY: requirements
 requirements: requirements.txt
-requirements.txt:
+requirements.txt: setup.py
        echo "" > requirements.txt
        # See also: https://github.com/peterbe/hashin/issues/139
        for pkg in `pip list --format freeze`; do hashin --verbose $$pkg; done
@@ -86,16 +86,16 @@ test:
 .PHONY: dist bdist-wheel sdist
 dist: bdist-wheel sdist
 bdist-wheel: $(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION)-py3-none-any.whl
-$(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION)-py3-none-any.whl:
+$(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION)-py3-none-any.whl: check test
        python setup.py bdist_wheel --dist-dir $(VIRTUAL_ENV)/dist/ --bdist-dir $(VIRTUAL_ENV)/build
 sdist: $(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION).tar.gz
-$(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION).tar.gz:
+$(VIRTUAL_ENV)/dist/package-$(PACKAGE_VERSION).tar.gz: check test
        python setup.py sdist --dist-dir $(VIRTUAL_ENV)/dist/
 
 # Build the HTML documentation from the package's source.
 .PHONY: docs
 docs: docs/_build/html/index.html
-docs/_build/html/index.html:
+docs/_build/html/index.html: check test
        $(MAKE) -C docs/ html
 
 # Clean test caches and remove build artifacts.
```

Makefile
```diff
@@ -0,0 +1,118 @@
...
+check:
+	pre-commit run --all-files
+
+# Run all unit tests.
+.PHONY: test
+test:
+	pre-commit run pytest --hook-stage push
+
+# Build a source distribution package and a binary wheel distribution artifact.
+.PHONY: dist bdist-wheel sdist
```
behnazh, line 86: So the assumption is that `VIRTUAL_ENV` env variable is set (to `.` in the Actions workflows for example). If empty, this goal will attempt creating the release artifact at `/dist/` which does not exist.

jenstroeger: Correct, some goals assume that `VIRTUAL_ENV` exists and points at a sensible folder, which is why we have this guard:

https://github.com/jenstroeger/python-package-template/blob/e25b624aebfd083a38794bc970afa307093da0e0/Makefile#L21-L23

and the README mentions that:

https://github.com/jenstroeger/python-package-template/blob/e25b624aebfd083a38794bc970afa307093da0e0/UPSTREAM_README.md#L107

If we were to run the `dist` goal as part of an Action then it would fail because the `VIRTUAL_ENV` variable doesn’t exist and we haven’t created a venv in an Action either (see https://github.com/jenstroeger/python-package-template/pull/74#issuecomment-1073001587, and issue https://github.com/actions/setup-python/issues/359). Because currently we only run the `setup`, `check` and `test` goals we’re ok without that variable.

jenstroeger: @behnazh so… you’re saying to remove `$(VIRTUAL_ENV)` entirely and use `dist/` and `build/` in the root folder, next to this Makefile?

Makefile
```diff
@@ -0,0 +1,118 @@
...
+# Remove code caches, or the entire virtual environment.
+.PHONY: nuke-caches nuke
+nuke-caches: clean
+	find src/ -name __pycache__ -exec rm -fr {} +
+	find tests/ -name __pycache__ -exec rm -fr {} +
+nuke: nuke-caches
+	if [ ! -z "${VIRTUAL_ENV}" ]; then echo "Nuking activated virtual environment!"; fi
+	rm -fr src/package.egg-info
+	rm -fr $(VIRTUAL_ENV)
+	rm -fr requirements.txt
```
jenstroeger, line 118: We should change to:
```makefile
nuke: nuke-caches
	if [ ! -z "${VIRTUAL_ENV}" ]; then \
	  echo "Nuking activated virtual environment!"; \
	  rm -fr $(VIRTUAL_ENV); \
	fi
	rm -fr src/package.egg-info
	rm -fr requirements.txt
```

behnazh:
> @behnazh, I wonder if we should create the following dependencies for some of the targets:

The dependencies for these targets make sense.

> We’ll need a $(VIRTUAL_ENV)/bin/.updated or some such timestamp, though, for the make upgrade dependency.
I didn't understand the point of a timestamp.

Makefile
behnazh: > @behnazh so… you’re saying to remove `$(VIRTUAL_ENV)` entirely and use `dist/` and `build/` in the root folder, next to this Makefile?

Yes, I believe we shouldn't fiddle with the virtual environment at all, and in environments like Actions, the release should be stored in `dist/` and not `.venv/dist/`.

It might be worthwhile adding a note to the README to show where the package will be stored if venv is activated?

jenstroeger:
> > We’ll need a $(VIRTUAL_ENV)/bin/.updated or some such timestamp, though, for the make upgrade dependency.
> 
> I didn't understand the point of a timestamp.

To avoid executing the recipe every time somebody types `make update`:

https://github.com/jenstroeger/python-package-template/blob/e25b624aebfd083a38794bc970afa307093da0e0/Makefile#L55-L57

The recipe (both `pip install …`) is executed if a target (`upgrade`) is out of date wrt. its prerequisites. But the above rule doesn’t have any prerequisites so it’s considered “out of date” with every run, thus both `pip install …` run with every `make upgrade` even though there’s nothing to upgrade ([docs](https://www.gnu.org/software/make/manual/html_node/Rule-Syntax.html)).

But if we specify a target:
```Makefile
upgrade: $(VIRTUAL_ENV)/bin/.upgraded
$(VIRTUAL_ENV)/bin/.upgraded:
	pip install --upgrade pip 
	pip install --editable .[hooks,dev,test,docs] 
	touch $(VIRTUAL_ENV)/bin/.upgraded
force-upgrade:
	rm -fr $(VIRTUAL_ENV)/bin/.upgraded
	$(MAKE) upgrade
```

We can then add that file as a prerequisite to

https://github.com/jenstroeger/python-package-template/blob/e25b624aebfd083a38794bc970afa307093da0e0/Makefile#L63

so that the `requirements.txt` is generated only if that `.updated` file is more recent. Or maybe that’s all making it a little too complicated 🤓

Makefile
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/198baff20711964fa6e45b3df63c9cc96bcb57ee.

Makefile
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/063f6eb441d06d350b88f2cb9a0797e2c6e9c34c.

jenstroeger:
I wonder if…

- `make dist` should depend on `check` and `test` because I wouldn’t want to build a distribution if the package doesn’t pass the checks and tests; and
- `make docs` same because I wouldn’t want to build documentation if the package doesn’t pass the checks and tests.

jenstroeger:
> https://github.com/jenstroeger/python-package-template/pull/74#issuecomment-1080217305
>
> The dependencies for these targets make sense. 

Commit https://github.com/jenstroeger/python-package-template/pull/74/commits/bc0e9de5eadba186dbf6601dc7456329a9a3ba7c.

behnazh (approved):

PR #90, opened 2022-01-06
chore: make sure to always show certain warnings when running tox

jenstroeger:
I think it makes sense to _always_ show certain runtime exceptions, e.g. a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). We may want to document this, and we may consider using the [-W](https://docs.python.org/3/using/cmdline.html#cmdoption-W) option to raise exceptions and actually fail a `tox` run?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ python =
 # Use pytest for testing and set up an optional environment for it.
 [testenv]
 extras = test
-commands = pytest
+commands = python -W always::DeprecationWarning -m pytest
 setenv =
     SOME_VAR = "value"
 


behnazh:
> I think it makes sense to _always_ show certain runtime exceptions, e.g. a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). We may want to document this, and we may consider using the [-W](https://docs.python.org/3/using/cmdline.html#cmdoption-W) option to raise exceptions and actually fail a `tox` run?

Sounds good!

jenstroeger:
> Sounds good!

Document it, or change `always` to `error`?

jenstroeger:
@behnazh same failure as #97 — do we now have to specify `src/package`? Oddly enough, this works
```
> mypy src
Success: no issues found in 3 source files
```

behnazh (approved):

behnazh:
> Shall we roll back `mypy` to its previous version [v0.930](https://github.com/python/mypy/releases/tag/v0.930)?

No, I don't see it necessary to roll back if we can adjust the configuration. How about passing the `-p package` option to `mypy`?

PR #91, opened 2022-01-06
ci: add more pre-commit hooks, and clean up what the hooks flagged

jenstroeger:
I added a bunch of new pre-commit hooks, and they reformatted the YAML and TOML files and found a few minor issues. Note, though, that said reformatting removed comments…

Bummer that I still haven’t figured out why three hooks run on `git push` when only one is configured to do so:
```
> git push --set-upstream origin pre-commit-improvements
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
tox......................................................................Passed
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,24 +1,26 @@
-# https://pycqa.github.io/isort/
+[tool.bandit]
+tests = []
+skips = ["B101"]
+
+[tool.black]
+line-length = 120
+target-version = ["py39"]
+
+[tool.coverage.report]
+fail_under = 100
+show_missing = true
+
+[tool.coverage.run]
+omit = [
+    "*/package/__main__.py",
+]
+
 [tool.isort]
 profile = "black"
 multi_line_output = 3
 line_length = 120
 skip_gitignore = true
 
-# https://github.com/psf/black#configuration
-[tool.black]
-line-length = 120
-target-version = ["py39"]
-
-# https://bandit.readthedocs.io/en/latest/config.html
-# Skip test B101 because of issue https://github.com/PyCQA/bandit/issues/457
-[tool.bandit]
-tests = []
-skips = ["B101"]
-
-# https://docs.pytest.org/en/6.2.x/customize.html#configuration-file-formats
-# https://docs.pytest.org/en/6.2.x/reference.html#configuration-options
-# https://docs.pytest.org/en/6.2.x/reference.html#command-line-flags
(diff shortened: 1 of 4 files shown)


jenstroeger:
> Bummer that I still haven’t figured out why three hooks run on `git push` when only one is configured to do so:
> 
> ```
> > git push --set-upstream origin pre-commit-improvements
> Fix End of Files.........................................................Passed
> Trim Trailing Whitespace.................................................Passed
> tox......................................................................Passed
> ```

So, the [default configuration](https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-hooks.yaml) for these hooks contains
```yaml
    stages: [commit, push, manual]
```
which is why the above hooks run on commit _and_ push. Fixed to run on commit only. I also added a `name` to most hooks to print a more readable output. Lastly, I removed language specific versions from the hooks to give them more flexibility wrt. to the hosting venv.

behnazh (approved):

pyproject.toml
```diff
@@ -1,24 +1,26 @@
-# https://pycqa.github.io/isort/
-[tool.isort]
-profile = "black"
-multi_line_output = 3
-line_length = 120
-skip_gitignore = true
+[tool.bandit]
```
behnazh, line 1: I don't think it's a good idea to let the linters remove the comments.

jenstroeger: I agree: the `pretty-format-toml` hook removes comments unless they’re at the top of the file.

Looking at [the code](https://github.com/macisamuele/language-formatters-pre-commit-hooks/blob/16762a19ddaa3a966ca7c4edd9ade6189f0df627/language_formatters_pre_commit_hooks/pretty_format_toml.py#L32) it uses the [toml_sort](https://github.com/pappasam/toml-sort) package which has an open issue https://github.com/pappasam/toml-sort/issues/11.

behnazh: How about removing `pretty-format-toml` until the issue is resolved and use a different hook?

jenstroeger: Considering that that hook does nothing but sorting the entries, we could add the comments back in and comment out the hook 👍🏼

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/91/commits/f8d4e61eed5c8cafa7ac3cd31d94c851ac13d064.

PR #109, opened 2022-01-20
chore: improve `mypy` error messages

jenstroeger:
This change adds [mypy error codes](https://mypy.readthedocs.io/en/stable/error_codes.html) to the error messages, which are then used to ignore only those messages; required by the `python-check-blanket-type-ignore` commit hook.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -10,6 +10,7 @@ mypy_path=src
 
 # Useful settings.
 show_column_numbers=True
+show_error_codes=True
 check_untyped_defs=True
 
 # And some more warnings.


mypy.ini
```diff
@@ -10,6 +10,7 @@ mypy_path=src
 
 # Useful settings.
 show_column_numbers=True
+show_error_codes=True
```
behnazh, line 13: Can you please add a comment why turning this option on is useful?

jenstroeger: Should we then explain the other options in this file as well?

behnazh: Maybe if they are interesting like this one :wink:

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/109/commits/4a3b0fbf8df2ad84c9d4fdb82e1a15d4b2b9a0d4.

behnazh (approved):

PR #114, opened 2022-01-29
chore: add “bump” type to release commits

openbrian:
commit_subject for semantic-release version command must have proper convention

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,5 +43,5 @@ testpaths = [
 branch = "main"
 version_variable = "src/package/__init__.py:__version__"
 upload_to_pypi = false
-commit_subject = "chore: Release v{version}"
+commit_subject = "bump: release v{version}"
 changelog_file = "CHANGELOG.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,5 +43,5 @@ testpaths = [
 branch = "main"
 version_variable = "src/package/__init__.py:__version__"
 upload_to_pypi = false
-commit_subject = "Release v{version}"
+commit_subject = "chore: Release v{version}"
 changelog_file = "CHANGELOG.md"


jenstroeger:
@openbrian, thank you for your contribution!

You make a good point, and at the moment I don’t have a good answer. The [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) doesn’t mention explicitly how to type a release commit, and I didn’t see it [discussed](https://github.com/conventional-commits/conventionalcommits.org/issues) either.

Interestingly, the [`python-semantic-release` package](https://github.com/relekang/python-semantic-release/commits/master) itself doesn’t type release commits, and the [Angular releases](https://github.com/angular/angular/commit/b2a081a19d6d006023b0f79076e9e499441a7810) (where conventional commits originate) are typed with `release` which I _didn’t_ see mentioned in the spec.

I wonder if this is a question worth raising with the folks who maintain the spec? In the meantime, I’m ok with typing release commits as “chore” (or “ci” or “build”) — with lower-case “release” for consistency’s sake:
```toml
commit_subject = "chore: release v{version}"
```

openbrian:
Maybe I'm not doing versioning right.  I want to handle version/changing separate from git and releases.  So I run git manually, and pre-commit catches this mistake.  I suspect python-semantic-release does a no verify commit.

jenstroeger:
Hmm. So, when the `semantic-release` tool runs as an Action on the Github side ([here](https://github.com/jenstroeger/python-package-template/blob/1d7ab62da35cf2e211bd154832aa33293bde5cb5/.github/workflows/check-and-release-main.yaml#L35-L47) is the setup for that) then it runs without any git hooks checking anything — the Github Action that executes `semantic-release` commits the release commit straight into the `main` branch on the server, using the supplied settings.

If you [run the `semantic-release` tool locally to create a release commit](https://python-semantic-release.readthedocs.io/en/latest/#semantic-release-version) then that’s just local, you need to push to your git remote yourself. I think that’s what you described above?

The change you propose in this PR affects both approaches — running `semantic-release` as a Github Action on the server and running the tool locally.

Is that what you mean?

relekang:
>Interestingly, the python-semantic-release package itself doesn’t type release commits (perhaps @relekang can shed some light on that decision), and the Angular releases (where conventional commits originate) are typed with release which I didn’t see mentioned in the spec.

That decision was made way to long ago to remember 😅

behnazh:
> I want to handle version/changing separate from git and releases. So I run git manually, and pre-commit catches this mistake.

I think it would be a valid usecase to manually create a release in certain circumstances, and as you pointed out the `commitizen` pre-commit hook fails on [the current release commit message](https://github.com/jenstroeger/python-package-template/blob/main/pyproject.toml#L46). 

I noticed that `commitizen` [also accepts](https://github.com/commitizen-tools/commitizen/blob/v2.20.0/commitizen/cz/conventional_commits/conventional_commits.py#L193) `bump` as the type. How about using `bump` as the type?

```
commit_subject = "bump: release v{version}"
```

openbrian:
I like bump.

jenstroeger:
Alas, the action we use for PR subject linting

https://github.com/jenstroeger/python-package-template/blob/1d7ab62da35cf2e211bd154832aa33293bde5cb5/.github/workflows/pull-request.yaml#L17-L26

is based on [commitlint](https://github.com/conventional-changelog/commitlint) which, [it seems](https://github.com/conventional-changelog/commitlint/blob/99d8881d0d951deded6d9e31bbb279d04101549b/%40commitlint/config-conventional/index.js#L18-L34), doesn’t accept `bump` or `release` as a type. I didn’t find a related issue either. That’s probably irrelevant though, because the Action commits directly to the `main` branch.

Similarly, the [python-semantic-release](https://github.com/relekang/python-semantic-release) package [doesn’t know](https://github.com/relekang/python-semantic-release/blob/3e34f957ff5a3ec6e6f984cc4a79a38ce4391ea9/semantic_release/defaults.cfg#L2) about `bump` or `release`, and I’ve not read through the code yet to see how a `bump: bla` or `release: bla` commit would affect the `semantic-release` tool…

behnazh:
>  the action we use for PR subject linting

I don't that matters in this case because the release commit is not a PR anyway.

> and I’ve not read through the code yet to see how a bump: bla or release: bla commit would affect the semantic-release tool

I've checked it manually and `python-semantic-release` successfully creates the release using `bump` as the type.

jenstroeger:
> I've checked it manually and `python-semantic-release` successfully creates the release using `bump` as the type.

…and it ignores existing `bump` commits, i.e. previous releases?

behnazh:
> …and it ignores existing `bump` commits, i.e. previous releases?

Hm, haven't checked if it affects the changelog in the next release.

jenstroeger:
> @behnazh: I noticed that `commitizen` [also accepts](https://github.com/commitizen-tools/commitizen/blob/v2.20.0/commitizen/cz/conventional_commits/conventional_commits.py#L193) `bump` as the type. How about using `bump` as the type?
> 
> ```
> commit_subject = "bump: release v{version}"
> ```

Hmph… but they don’t export that in their [conventional-commit-types](https://github.com/commitizen/conventional-commit-types) package [here](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) 🤔

behnazh:
> Hmph… but they don’t export that in their [conventional-commit-types](https://github.com/commitizen/conventional-commit-types) package [here](https://github.com/commitizen/conventional-commit-types/blob/master/index.json), and neither is it part of [their choice list](https://github.com/commitizen-tools/commitizen/blob/v2.20.0/commitizen/cz/conventional_commits/conventional_commits.py#L48-L109) thinking

I'm not sure where/how/if [index.json](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) is used? And I think this should be discussed in [conventional-commit-types](https://github.com/commitizen/conventional-commit-types) because in practice`bump` works.

jenstroeger:
> I'm not sure where/how/if [index.json](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) is used?

I think there are some references under the [Etc](https://github.com/commitizen/conventional-commit-types#etc) section.

> And I think this should be discussed in [conventional-commit-types](https://github.com/commitizen/conventional-commit-types) because in practice `bump` works.

I agree; I’d even go as far as raising this whole issue with the [conventionalcommit.org](https://github.com/conventional-commits/conventionalcommits.org) repo. And once it’s resolved there — one way or another — these tools should pick up the change, if there is any.

And for this repo? Shall we go with `bump` as a default type for release commits for now?

behnazh:
> > I'm not sure where/how/if [index.json](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) is used?
> 
> I think there are some references under the [Etc](https://github.com/commitizen/conventional-commit-types#etc) section.
> 
> > And I think this should be discussed in [conventional-commit-types](https://github.com/commitizen/conventional-commit-types) because in practice `bump` works.
> 
> I agree; I’d even go as far as raising this whole issue with the [conventionalcommit.org](https://github.com/conventional-commits/conventionalcommits.org) repo. And once it’s resolved there — one way or another — these tools should pick up the change, if there is any.
> 
> And for this repo? Shall we go with `bump` as a default type for release commits for now?

Yes :+1:

pyproject.toml
```diff
@@ -43,5 +43,5 @@ testpaths = [
 branch = "main"
 version_variable = "src/package/__init__.py:__version__"
 upload_to_pypi = false
-commit_subject = "Release v{version}"
+commit_subject = "chore: Release v{version}"
```
jenstroeger, line 46: As per the above discussion, let’s change this line to
```toml
commit_subject = "bump: release v{version}"
```

jenstroeger (changes requested):
@openbrian, thank you for kicking off this conversation! Please see the feedback below.

Can you also please change the PR title accordingly, and make sure the title itself follows conventional commit rules?

jenstroeger (approved):

jenstroeger:
@relekang if you’re interested then I’d open a PR for your [python-semantic-release](https://github.com/relekang/python-semantic-release) package which adds `bump:` to [the default](https://github.com/relekang/python-semantic-release/blob/c59095ed5953af2efd418ac7aec772e880b3ece3/semantic_release/defaults.cfg#L18), and document accordingly?

relekang:
Yes, sounds good, needs to be a major release tho, since someone might be relying on the current commit messages in their ci pipelines.

behnazh (approved):

PR #105, opened 2022-01-18
feat: add scorecards analysis workflow

behnazh:
This PR adds [Scorecards security Actions workflow](https://github.com/ossf/scorecard-action), which analyzes the repo for various supply-chain security issues.

Closes #32

PR #140, opened 2022-03-04
ci: add Release Notification action

jenstroeger:
This change adds a new Github Action for the [Release Event](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#releaseevent): when a release is published (either manually through the Github website or through another Action) then this action sends out a Slack notification.

Because we don’t actually have a Slack workspace with this repository, the Action’s `slack` job is being skipped:

https://github.com/jenstroeger/python-package-template/blob/06319c75a537cf6df8d98bb71637c79b93e7ced0/.github/workflows/release-notifications.yaml#L8-L10

For now this PR is a **Draft** because we should probably document this feature in the [README.md](https://github.com/jenstroeger/python-package-template/blob/main/UPSTREAM_README.md)…

### DIFF
--- a/.github/workflows/release-notifications.yaml
+++ b/.github/workflows/release-notifications.yaml
@@ -1,0 +1,30 @@
+name: Release Notifications
+on:
+  release:
+    types: [published]
+jobs:
+  slack:
+    name: Slack release notification
+    # Remove the next line after your Slack app has been set up. See also:
+    # https://github.com/slackapi/slack-github-action#setup-2
+    if: ${{ false }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: slackapi/slack-github-action@v1.18.0
+      with:
+        payload: |
+          {
+            "text": "${{ github.event.repository.name }} published a new release ${{ github.event.release.tag_name }}",
+            "blocks": [
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "*${{ github.event.repository.name }}* published a new release <${{ github.event.release.html_url }}|${{ github.event.release.tag_name }}>"
+                }
+              }
+            ]
+          }
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
--- a/.github/workflows/release-notifications.yaml
+++ b/.github/workflows/release-notifications.yaml
@@ -2,15 +2,21 @@ name: Release Notifications
 on:
   release:
     types: [published]
+
+# Declare default permissions as read only.
+permissions: read-all
+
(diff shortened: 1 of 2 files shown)


behnazh (changes requested):
That's really cool.

Please add `permissions: read-all` at the top of the workflow and grant necessary permissions per `job`. Also the third-party workflows need to be pinned. You can follow [this](https://github.com/jenstroeger/python-package-template/pull/105/files#diff-5fa68b2c6ae56bb9d97abf7dc1ffbfc79e99b6404338daa86648b5a00764b9f4) workflow as an example.

jenstroeger:
BTW, I checked: both the [`gh` CLI client](https://cli.github.com/) and the [REST API](https://docs.github.com/en/rest/reference/releases#create-a-release) (e.g. using [curl](https://curl.se/)) create a proper release and trigger the [ReleaseEvent](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#releaseevent) correctly.

So, my suspicion for the event not being triggered the other day (i.e. the Action didn’t run and therefore no Slack notification) is that the Action was merged into the tree with the same commit that ran the workflow. That didn’t prevent Actions from running in the past but it’s the only explanation I have right now. I’ll wait for the next release and if _again_ the Action doesn’t run then I’ll dig around more.

jenstroeger:
> Please add `permissions: read-all` at the top of the workflow and grant necessary permissions per `job`. Also the third-party workflows need to be pinned. You can follow [this](https://github.com/jenstroeger/python-package-template/pull/105/files#diff-5fa68b2c6ae56bb9d97abf7dc1ffbfc79e99b6404338daa86648b5a00764b9f4) workflow as an example.

@behnazh, done. Is the permission here

https://github.com/jenstroeger/python-package-template/blob/06e83a8741bd51d3dc3f27f1091c381ad639949e/.github/workflows/release-notifications.yaml#L15-L16

necessary considering

https://github.com/jenstroeger/python-package-template/blob/06e83a8741bd51d3dc3f27f1091c381ad639949e/.github/workflows/release-notifications.yaml#L6-L7

Also, would extra paranoia make sense? I mean, set
```yaml
# Declare default permissions as read only.
permissions: none-all
```
and enable read access only for the job
```yaml
    permissions:
      contents: read
```

behnazh:
> On second thought, the Action might need no permissions at all because it takes all information from `ReleaseEvent` only?

It looks like it because we don't even need to `checkout` the source code. So, maybe change it to `permissions: none-all` and see what happens?

jenstroeger:
> It looks like it because we don't even need to `checkout` the source code. So, maybe change it to `permissions: none-all` and see what happens?

Works, commit https://github.com/jenstroeger/python-package-template/pull/140/commits/dcc5e043e8ebeda70b661f7bc101bee82583435d.

behnazh (approved):

PR #183, opened 2022-04-26
chore: adjust Dependabot configuration

behnazh:
Changes the Dependabot configuration to run the checks weekly instead of daily. Also changes the `fix` commit message scopes to `chore`.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,9 +7,9 @@ updates:
 - package-ecosystem: pip
   directory: /
   schedule:
-    interval: daily
+    interval: weekly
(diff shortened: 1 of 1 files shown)


jenstroeger (changes requested):
Looks good, except the updated `src/package/__init__.py` file 😉

behnazh:
> Looks good, except the updated `src/package/__init__.py` file wink

That happened because `staging` is not rebased on `main` I think.

jenstroeger:
You’re right, done!

jenstroeger (approved):

PR #190, opened 2022-05-05
fix: fix "make dist" failing on latest version of Ubuntu

behnazh:
This PR installs the `wheel` package which is necessary for running `python setup.py bdist_wheel` ([make dist](https://github.com/jenstroeger/python-package-template/blob/add-wheel/Makefile#L100)) on latest version of Ubuntu. It also adds new steps to the CI workflow to test the the build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ setuptools.setup(
             "pep8-naming==0.12.1",
             "pylint>=2.9.3,<=2.13.8",
             "types-setuptools>=57.4.7,<=57.4.14",
+            "wheel==0.37.1",
         ],
         "docs": [
             "sphinx>=4.1.2,<=4.5.0",
--- a/.github/workflows/check-and-release-main.yaml
+++ b/.github/workflows/check-and-release-main.yaml
@@ -28,6 +28,10 @@ jobs:
       run: make check
     - name: Run tests
       run: make test
+    - name: Build the package
+      run: make dist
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -77,6 +77,7 @@
             "pep8-naming==0.12.1",
             "pylint>=2.9.3,<=2.13.8",
             "types-setuptools>=57.4.7,<=57.4.14",
+            "wheel>=0.37.1",
```
jenstroeger, line 80: I think we should either pin to
```python
"wheel==0.37.1"
```
or constrain the next-up version to avoid breaking changes:
```python
"wheel>=0.37.1,<1.0.0"
```

behnazh: Sure, I will pin it to `wheel==0.37.1`.

jenstroeger (commented):
I’m tempted to make the `build.yaml` part of a more general Actions cleanup. Also, if we’re testing the build then maybe it makes sense to add `make docs` as well?

jenstroeger (approved):

jenstroeger:
@behnazh I just realized that we should’ve also updated [pull-request.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/pull-request.yaml) 🤦🏻‍♂️

behnazh:
> I am _**very**_ tempted to undo the last two commit to `main`.

ok

PR #208, opened 2022-05-25
feat: move package specification and tool configs to pyproject.toml

### LINKED ISSUE #202: Migrate from setup.py to pyproject.toml
We currently use a [setup.py](https://github.com/jenstroeger/python-package-template/blob/main/setup.py) file to store package metadata and installation information. However, as the [`pip` documentation mentions](https://pip.pypa.io/en/stable/reference/build-system/setup-py/):

> Prior to the introduction of pyproject.toml-based builds (in [PEP 517](https://peps.python.org/pep-0517/) and [PEP 518](https://peps.python.org/pep-0518/)), pip had only supported installing packages using setup.py files that were built using [setuptools](https://pypi.org/project/setuptools/). The interface documented here is retained currently solely for legacy purposes, until the migration to pyproject.toml-based builds can be completed.

And from issue https://github.com/slsa-framework/slsa-github-generator/issues/55:

> I think it's really our only option. We can do some things to guess the build backend that needs to be present, but it's not guaranteed to be correct, or capture all bulid-time dependencies. Instead, pyproject.toml is the standardized way to do this. I think it would be reasonable to say we only support building projects that specify pyproject.toml.

Let’s move on then, and move to [pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/). For inspiration, @hynek has moved the [**structlog** package to pyproject.toml](https://github.com/hynek/structlog/blob/main/pyproject.toml) (and uses [Flit](https://flit.pypa.io/) instead of [Setuptools](https://setuptools.pypa.io/en/latest/)).

And while we’re at it:
- Let’s [merge](https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml) the [mypy.ini](https://github.com/jenstroeger/python-package-template/blob/main/mypy.ini) file into pyproject.toml as well?
- Let’s [merge](https://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options) the [pylintrc](https://github.com/jenstroeger/python-package-template/blob/main/pylintrc) file into pyproject.toml as well?
- Unfortunately, `flake8` does not (yet?) support pyproject.toml (see also issue https://github.com/PyCQA/flake8/issues/234).
- Unfortunately, our PR Commit Lint Action does not (yet?) support pyproject.toml (see also issue https://github.com/CondeNast/conventional-pull-request-action/issues/15).

jenstroeger:
Closes #202, and for more details see that issue.

Three thoughts:
- ~Update the README according to these changes.~
- ~We should remove the `hypothesis` package dependency from the `"dev"` optional deps.~
- ~I think the following overrides should be moved into the main section of the mypy configuration:~ https://github.com/jenstroeger/python-package-template/blob/b0d802a6b95225a238b1f0dc78ca4757d549a46d/pyproject.toml#L113-L118


Makefile
```diff
@@ -91,19 +91,19 @@ test:
...
 else
   PACKAGE_VERSION=$(shell python -c 'import package; print(package.__version__)')
 endif
-dist: bdist-wheel sdist
-bdist-wheel: dist/package-$(PACKAGE_VERSION)-py3-none-any.whl
+dist: wheel sdist
+wheel: dist/package-$(PACKAGE_VERSION)-py3-none-any.whl
 dist/package-$(PACKAGE_VERSION)-py3-none-any.whl: check test
-	python setup.py bdist_wheel --dist-dir dist/ --bdist-dir build/
+	flit build --setup-py --format wheel
```
behnazh, line 103: Not sure if we would need to set `SOURCE_DATE_EPOCH=$(date +%s)` to avoid non-deterministic time stamps for [reproducible builds](https://flit.pypa.io/en/latest/reproducible.html)? Or is it just effective when we publish using `flit`?

jenstroeger: My interpretation is that that env variable is for `publish` only 🤔

behnazh: I think we should open a separate issue for reproducible builds.

pyproject.toml
```diff
@@ -20,13 +85,64 @@ omit = [
...
+[tool.pylint.MASTER]
+fail-under = 10.0
+disable = [
+    "fixme",
+    "too-many-statements",
+    "too-many-branches",
+    "too-many-locals",
+    "too-many-nested-blocks",
+    "too-many-return-statements",
+    "too-few-public-methods",
```
behnazh, line 131: How about these two other `too-many` warnings?

- `too-many-arguments`
- `too-many-instance-attributes`

Makefile
jenstroeger: Agreed, and probably with @relekang’s [python-semantic-release](https://github.com/relekang/python-semantic-release) repo, independent of this one.

pyproject.toml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/208/commits/c1bf2ddd78761b27ae2eb5f107db3a1bafeb6d7d.

Maybe we should review [all messages](https://pylint.pycqa.org/en/latest/user_guide/messages/messages_overview.html) to make sure we’ve got the `too-many-*` and `too-few-*` covered?

behnazh (approved):

pyproject.toml
jenstroeger: Naw, the other `too-many-*` kinda make more sense… let’s leave it at what we have.

PR #209, opened 2022-05-27
fix(ci): enable Release Action for protected branches

jenstroeger:
Thanks you @lecram for looking into this!

### DIFF
--- a/.github/workflows/check-and-release-main.yaml
+++ b/.github/workflows/check-and-release-main.yaml
@@ -54,7 +54,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
       with:
-        # For protected branches, see: https://github.com/relekang/python-semantic-release/issues/311
+        # Private repos with protected branches may need to set that token to make the Action work. For a
+        # discussion of this problem, see: https://github.com/relekang/python-semantic-release/issues/311
         token: ${{ secrets.GITHUB_TOKEN }}
         fetch-depth: 0
     - name: Python Semantic Release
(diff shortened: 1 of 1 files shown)


.github/workflows/check-and-release-main.yaml
```diff
@@ -51,9 +51,11 @@ jobs:
       uses: step-security/harden-runner@248ae51c2e8cc9622ecf50685c8bf7150c6e8813
       with:
         egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
-
-    - uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
+    - name: Checkout
+      uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
       with:
+        # For protected branches, see: https://github.com/relekang/python-semantic-release/issues/311
+        token: ${{ secrets.GITHUB_TOKEN }}
```
behnazh, line 58: Is this really necessary for this repo? The `token` secret should be available by default in `checkout@v3` according to the documentation [here](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token).

jenstroeger: I added this for documentation’s sake with a reference to a related issue. Private repos with protected branches may need to set that token to make the Action work.

behnazh: > I added this for documentation’s sake with a reference to a related issue. Private repos with protected branches may need to set that token to make the Action work.

How about adding this to the comment?

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/209/commits/0fd24a03caaa718786a718220bc8ca0319ea0495.

behnazh (approved):

PR #211, opened 2022-06-02
refactor(ci): refactor and address security issues in workflows

behnazh:
This PR:

- refactors the Actions workflows by adding `build.yaml` as a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to check, test, and build the package. This workflow can be called from other workflows and avoids copy pasting the build code.
- fixes the [Dangerous-Workflow](https://github.com/jenstroeger/python-package-template/security/code-scanning/29) and [Token-Permission](https://github.com/jenstroeger/python-package-template/security/code-scanning/23) issues reported by Scorecards.
- bumps the Scorecard Actions workflow to `1.1.1`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,5 +157,5 @@ testpaths = [
 branch = "main"
 version_variable = "src/package/__init__.py:__version__"
 upload_to_pypi = false
-commit_subject = "bump: release v{version}"
+commit_subject = "bump: release v{version} [skip actions]"
 changelog_file = "CHANGELOG.md"


.github/workflows/build.yaml
```diff
@@ -0,0 +1,41 @@
...
+  contents: read
+
+jobs:
+  build:
+    name: Check Python ${{ matrix.python }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
jenstroeger, line 17: Should we leave a comment that it’s recommended to [pin a Runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) version specifically?

jenstroeger (approved):
Looks good 👍🏼  

One comment: [conventional commits](https://www.conventionalcommits.org/) do have a `refactor(ci):` type so we don’t have to use `chore:` here.

.github/workflows/build.yaml
behnazh: See commit 8db20dd4e6538c089f77e3aa9c4dcbf0b53ee422.

.github/workflows/pull-request.yaml
```diff
@@ -26,38 +26,12 @@ jobs:
       with:
         egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
 
-    - uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
```
behnazh, line 29: @jenstroeger by removing the `checkout` step here, we get a [warning](https://github.com/jenstroeger/python-package-template/runs/6717525442?check_suite_focus=true#step:4:9): `rules module not found` because it can't read [commitlint.rules.js](https://github.com/jenstroeger/python-package-template/blob/main/commitlint.rules.js).

We could as well remove the `commitlint.rules.js` file if we are happy with this warning?

jenstroeger: Hm, looking at the [default settings](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional) the `commitlint.rules.js` we have changes [`body-max-line-length`](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#body-max-line-length) from 100 to 200. Also, unfortunately, the [Conventional Commit Action](https://github.com/CondeNast/conventional-pull-request-action) doesn’t support passing these config values through the command line either.

So I guess yes, we might as well remove that file…

behnazh: see [3d6a52d](https://github.com/jenstroeger/python-package-template/pull/211/commits/3d6a52d212bae6c362219cdf2a8a7193ee3248a6).

.github/workflows/build.yaml
```diff
@@ -0,0 +1,43 @@
...
+        python: ['3.9', '3.10']
+    steps:
+    - name: Harden Runner
+      uses: step-security/harden-runner@248ae51c2e8cc9622ecf50685c8bf7150c6e8813
+      with:
+        egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+
+    - uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
+    - name: Set up Python
+      uses: actions/setup-python@98f2ad02fd48d057ee3b4d4f66525b231c3e52b6
```
jenstroeger, line 29: Should we update this to commit `d09bd5e6005b175076f227b13d9730d56e9dcfcb` (v4) of the Action — see also PR https://github.com/jenstroeger/python-package-template/pull/212?

jenstroeger (commented):
See also PR https://github.com/jenstroeger/python-package-template/pull/205/ which we can add to this change?

jenstroeger:
And while we’re at it, these entries need a `name:` to be consistent with the rest of the job names:

https://github.com/jenstroeger/python-package-template/blob/3d6a52d212bae6c362219cdf2a8a7193ee3248a6/.github/workflows/build.yaml#L27

https://github.com/jenstroeger/python-package-template/blob/3d6a52d212bae6c362219cdf2a8a7193ee3248a6/.github/workflows/pull-request.yaml#L29

https://github.com/jenstroeger/python-package-template/blob/3d6a52d212bae6c362219cdf2a8a7193ee3248a6/.github/workflows/release-notifications.yaml#L22

Also, should we keep the names in normal case or title case?

jenstroeger:
@behnazh I think we could remove these two lines:

https://github.com/jenstroeger/python-package-template/blob/66347c69af6dd9c30b2b85ebe022d2844573faee/.github/workflows/pull-request.yaml#L33-L35

PR #215, opened 2022-06-14
ci: add a bunch of useful Flake8 plugins to git commit hooks

jenstroeger:
Poking through a few other packages I came across these `flake8` plugins that I think are useful to add:
- [bugbear](https://github.com/PyCQA/flake8-bugbear)
- [comprehensions](https://github.com/adamchainz/flake8-comprehensions)
- [mutable](https://github.com/ebeweber/flake8-mutable) (somewhat dated though)
- [noqa](https://github.com/plinss/flake8-noqa)
- [pytest-style](https://github.com/m-burst/flake8-pytest-style)

We’d also consider:
- [scream](https://github.com/MartinThoma/flake8-scream)
- Review this list of [Awesome Flake8 Plugins](https://github.com/DmytroLitvinov/awesome-flake8-extensions)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -7,3 +7,7 @@ max-line-length = 120
 # Enable Bugbear's extended opinionated checks.
 # https://github.com/PyCQA/flake8-bugbear#how-to-enable-opinionated-warnings
 extend-select = B9
+
+# Ensure that flake8 warnings are silenced correctly.
+# https://github.com/plinss/flake8-noqa#options
+noqa-require-code = true
--- a/.flake8
+++ b/.flake8
@@ -1,8 +1,10 @@
 # Unfortunately, flake8 does not support pyproject.toml configuration.
 # https://github.com/PyCQA/flake8/issues/234
 [flake8]
-show-source = true
+ignore =
+per-file-ignores =
 max-line-length = 120
+show-source = true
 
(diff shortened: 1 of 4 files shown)


behnazh:
In general I would be careful about the quality, active maintenance, and security posture of these dependencies we pull in. With that in mind, do you still want to add all of these tools?

jenstroeger:
> In general I would be careful about the quality, active maintenance, and security posture of these dependencies we pull in. With that in mind, do you still want to add all of these tools?

I personally will enable these plugins to find out what they find out; they don’t modify any code anyway. Worst case is a lot of dismissible noise.

For this change to the Template Repository we can comment out all plugins and leave a comment for the user. I’m easy.

behnazh:
> they don’t modify any code anyway

I thought they might be.

I also find some of the checks interesting, e.g., the [comprehensions](https://github.com/adamchainz/flake8-comprehensions). Though I also wonder if there would be overlaps with Pylint? In general finding the right checkers that don't add too much noise will be important as the projects grow. Unfortunately this process is manual and not easy. 

Developers can comment out and remove the checkers that are not adding value selectively. I'm happy to approve.

behnazh (approved):

PR #239, opened 2022-07-01
feat!: replace the Makefile’s quick-check goal with check-code

jenstroeger:
A few new goals which, I think, make daily developer life a little easier.

If we consider the Makefile as an interface of this repo (which it kinda is) then this is a breaking change 🤔

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,15 @@ requirements.txt: pyproject.toml
 all: check test dist docs
 
 # Run some or all checks over the package code base.
-.PHONY: quick-check check
-quick-check:
+.PHONY: check check-code check-bandit check-flake8 check-lint check-mypy
+check-code: check-bandit check-flake8 check-lint check-mypy
+check-bandit:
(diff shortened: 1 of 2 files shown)


behnazh:
> If we consider the Makefile as an interface of this repo (which it kinda is) then this is a breaking change thinking

Yes, it seems to be a breaking change because the `quick-check` target is removed.

behnazh (approved):

jenstroeger:
We can also keep:
```makefile
quick-check: check-lint check-mypy
```
for compatibility’s sake. But, to be honest, I don’t think anybody cares about this ☺️

So… merge as breaking change, or just a simple `feat:`?

behnazh:
> We can also keep:
> 
> ```makefile
> quick-check: check-lint check-mypy
> ```

I think `quick-check` is vague and arbitrary anyway. So, I'm happy to remove it and bump as a breaking change.

behnazh (approved):

PR #244, opened 2022-07-05
fix: flit doesn’t support MANIFEST.in, fix sdist accordingly

jenstroeger:
Following up on our Slack conversation and issue https://github.com/pypa/flit/issues/565, this change fixes `flit`’s source distribution.

Interestingly installing the `.tar.gz` sdist file fails:
```
> pip install /path/to/template/dist/package-1.6.1.tar.gz 
Processing /path/to/template/dist/package-1.6.1.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [29 lines of output]
      Traceback (most recent call last):
        File "/.../T/pip-build-env-n3xxqt2l/overlay/lib/python3.9/site-packages/flit_core/config.py", line 273, in description_from_file
          with desc_path.open('r', encoding='utf-8') as f:
        File "/.../3.9/lib/python3.9/pathlib.py", line 1252, in open
          return io.open(self, mode, buffering, encoding, errors, newline,
        File "/.../3.9/lib/python3.9/pathlib.py", line 1120, in _opener
          return self._accessor.open(self, flags, mode)
      FileNotFoundError: [Errno 2] No such file or directory: 'README.md'
```
because the `README.md` is a symlink to the actual `UPSTREAM_README.md`. I think we should consider getting rid of that… (PR https://github.com/jenstroeger/python-package-template/pull/245)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,6 @@ exclude = [
     "CHANGELOG.md",
     "Makefile",
     "SECURITY.md",
-    "UPSTREAM_*.md",
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,24 @@ omit = [
 ]
 
 
+# https://flit.pypa.io/en/latest/pyproject_toml.html#sdist-section
+# See also: https://github.com/pypa/flit/issues/565
+[tool.flit.sdist]
+include = []
+exclude = [
+    ".github/",
+    "docs/",
+    "tests/",
+    ".flake8",
+    ".gitignore",
+    ".pre-commit-config.yaml",
+    "CHANGELOG.md",
+    "Makefile",
+    "SECURITY.md",
+    "UPSTREAM_*.md",
+]
+
+
 # https://pycqa.github.io/isort/
 [tool.isort]
 profile = "black"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -142,13 +142,6 @@ repos:
   # - id: pretty-format-toml
   #   args: [--autofix]
 
-# Make sure that the package's MANIFEST.in file works.
-- repo: https://github.com/mgedmin/check-manifest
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -84,6 +84,24 @@ omit = [
...
+    ".github/",
+    "docs/",
+    "tests/",
+    ".flake8",
+    ".gitignore",
+    ".pre-commit-config.yaml",
+    "CHANGELOG.md",
+    "Makefile",
+    "SECURITY.md",
+    "UPSTREAM_*.md",
```
behnazh, line 101: This might not to change when #245 is merged.

I see that you're keeping `README.md`. Wouldn't shipping `CHANGELOG.md` be a good idea too?

jenstroeger: > This might have to change when https://github.com/jenstroeger/python-package-template/pull/245 is merged.

Yup!

> I see that you're keeping `README.md`.

`flit` needed both `README.md` and `LICENSE.md` included so these two files can be added to the source dist.

> Wouldn't shipping `CHANGELOG.md` be a good idea too?

`flit` didn’t ask for it, so I haven’t added it yet. The changelog is also part of the metadata:

https://github.com/jenstroeger/python-package-template/blob/f8af74ddda359105680a06969de79382cf78f8ef/pyproject.toml#L58

I couldn’t find [mention of the changelog in sdist](https://docs.python.org/3/distutils/sourcedist.html#specifying-the-files-to-distribute), maybe @takluyver can shed some light on that?

jenstroeger: > This might not to change when #245 is merged.

Commit https://github.com/jenstroeger/python-package-template/pull/244/commits/adf530d938045c702b32e3fce478fa21cb05c52a.

behnazh (approved):

PR #248, opened 2022-07-07
feat: move private file .upgraded into .venv/ folder

### LINKED ISSUE #238: Unnecessary env in Action steps
I think these environment variables:

https://github.com/jenstroeger/python-package-template/blob/66347c69af6dd9c30b2b85ebe022d2844573faee/.github/workflows/build.yaml#L35-L36

https://github.com/jenstroeger/python-package-template/blob/66347c69af6dd9c30b2b85ebe022d2844573faee/.github/workflows/codeql-analysis.yaml#L51-L52

aren’t necessary because `make setup` assumes a venv. The `PYTHON` environment variable is used only by the `make venv` goal to pick the Python for the virtual environment:

https://github.com/jenstroeger/python-package-template/blob/66347c69af6dd9c30b2b85ebe022d2844573faee/Makefile#L29-L42

### LINKED ISSUE #240: Move `.upgraded` helper file into the virtual environment
Our Makefile uses a `.upgraded` file to indicate when/if the venv has been upgraded:

https://github.com/jenstroeger/python-package-template/blob/66347c69af6dd9c30b2b85ebe022d2844573faee/Makefile#L56-L60

I think the file should live in `${VIRTUAL_ENV}/.upgraded` instead of the repo’s root, because it’s concerned with the venv only and removing the venv folder manually wouldn’t leave behind that now-stale file.

jenstroeger:
Closes #238 #240 

And while at it, I renamed the file to `upgraded-on` and added a timestamp:
```
> cat .venv/upgraded-on 
Automatically generated by Python Package Makefile on 2022-07-07 14:13:19+10:00.
```
In order to make this work in CI, I also changed the Actions to create a virtual environment (see also https://github.com/actions/setup-python/issues/359#issuecomment-1080679055) such that our Actions workflows now operate on virtual environments using the _complete_ Makefile 👍🏼


behnazh (changes requested):
Please remove `.upgraded` from [.gitignore](https://github.com/jenstroeger/python-package-template/blob/move-upgraded/.gitignore#L137).

.github/workflows/codeql-analysis.yaml
```diff
@@ -46,11 +46,15 @@ jobs:
       uses: actions/setup-python@98f2ad02fd48d057ee3b4d4f66525b231c3e52b6
       with:
         python-version: ${{ matrix.python }}
```
behnazh, line 48: I think we can turn off CodeQL on `py 3.9`. CodeQL takes a lot of time and I'm not sure about the benefit of checking against different Python versions.

jenstroeger:
> Please remove `.upgraded` from [.gitignore](https://github.com/jenstroeger/python-package-template/blob/move-upgraded/.gitignore#L137).

Commit https://github.com/jenstroeger/python-package-template/pull/248/commits/aa64af7a2f96fb0cd6b0cfaef797baa025ae811e.

.github/workflows/codeql-analysis.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/248/commits/846ef236ea3fc0b54b93fd657972ff6e4b051c0d.

jenstroeger:
Commit https://github.com/jenstroeger/python-package-template/pull/248/commits/3e291762c3971eede030ad003b17414ee15a15f8 adds a check that compares the `pyproject.toml`’s last modification date with that of `.venv/upgraded-on`, and if the `.toml` is newer—indicating an update to the package metadata and dependencies—warns the user.

With that in mind, I think this PR becomes more of a `feat:` than a `chore:`?

behnazh:
> With that in mind, I think this PR becomes more of a `feat:` than a `chore:` or `refactor:`?

Yes, sounds good.

.github/workflows/build.yaml
```diff
@@ -30,10 +30,22 @@ jobs:
       uses: actions/setup-python@d09bd5e6005b175076f227b13d9730d56e9dcfcb
       with:
         python-version: ${{ matrix.python }}
-    - name: Install dependencies
-      run: make setup
+    - name: Create and activate virtual environment
+      if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest' }}
```
behnazh, line 34: Shouldn't this if statement be:
```
${{ matrix.os }} == 'ubuntu-latest' || ${{ matrix.os }} == 'macos-latest' 
```

.github/workflows/build.yaml
```diff
@@ -30,10 +30,22 @@ jobs:
...
+      run: |
+        make venv
+        source .venv/bin/activate
+      env:
+        PYTHON: python
+    - name: Create and activate virtual environment (Windows)
+      if: ${{ matrix.os == 'windows-latest' }}
```
behnazh, line 41: See my comment above.

.github/workflows/build.yaml
jenstroeger: Not according to original conversation on the Action: https://github.com/actions/setup-python/issues/359#issuecomment-1080679055

Also, the tests run successfully, so guess that code works 🤓

.github/workflows/build.yaml
jenstroeger: See my comment above.

.github/workflows/build.yaml
behnazh: You don't need to evaluate the [expression](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions) in `if` statements.

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. 

See this [example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository).

jenstroeger: It’s about consistency: we use `${{ … }}` in all workflows, so I wouldn’t want to change that here in this one instance just because, why?

PR #253, opened 2022-07-10
feat: warn if generated builds are not reproducible

jenstroeger:
For more details, see [this section](https://flit.pypa.io/en/latest/reproducible.html). Should we add a link to that document to the Makefile for reference?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,8 +95,8 @@ test:
 	pre-commit run pytest --hook-stage push
 
 # Build a source distribution package and a binary wheel distribution artifact.
-# When building these artifacts, we need the environment variable  SOURCE_DATE_EPOCH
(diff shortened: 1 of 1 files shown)


jenstroeger:
Hm, on second thought, instead of running that check as part of the recipe we could also run it as part of the Makefile itself, e.g. something like…
```makefile
ifneq ($(findstring dist,$(MAKECMDGOALS)),)
  ifeq ($(origin SOURCE_DATE_EPOCH),undefined)
    $(warning SOURCE_DATE_EPOCH variable not specified, building non-reproducible distribution)
  endif
endif
```

behnazh:
> For more details, see [this section](https://flit.pypa.io/en/latest/reproducible.html). Should we add a link to that document to the Makefile for reference?

Yes, I think so.

behnazh (approved):

behnazh (approved):

PR #255, opened 2022-07-10
fix: when nuking code and test caches, also nuke pre-commit hooks

jenstroeger:
From the [pre-commit docs](https://pre-commit.com/#pre-commit-clean):

> **pre-commit clean [options]**
> 
> Clean out cached pre-commit files.
> 
> Options: (no additional options)

So I thought that this would be useful when calling `make nuke-caches`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,7 @@ clean: dist-clean
 # Remove code caches, or the entire virtual environment.
 .PHONY: nuke-caches nuke
 nuke-caches: clean
+	pre-commit clean
 	find src/ -name __pycache__ -exec rm -fr {} +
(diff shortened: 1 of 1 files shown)


behnazh (commented):
I think we used to clean the `pre-commit` hooks  a while ago but decided to disable it because it cleans the hooks globally, which can have side effects on other projects.

jenstroeger:
Right, we also used to have it as part of `make clean` which was a much more common operation.

Close this PR?

behnazh:
> Close this PR?

Yes I think.

jenstroeger:
Because @behnazh said so.

PR #259, opened 2022-07-16
feat: generate SLSA level 3 provenance for release artifacts

### LINKED ISSUE #25: Prepare the repo to meet SLSA security levels
[SLSA](https://slsa.dev/) is a promising work that aims at improving the security posture of open source software and protecting against supply chain attacks. I believe this template provides a good opportunity to adopt SLSA. Even though SLSA is still a work in progress, I don't think we should wait for them to finalize the specs, and the earlier we start the better.

The current layout of the repo looks like this, which misses a build step (the quality assurance steps don't modify code, but build does)
image

To make this template and the repos created from it meet the highest [SLSA levels](https://slsa.dev/levels), we need to prepare the repo to meet their [requirements](https://slsa.dev/requirements). To begin with, I propose to make the repo ready for level 1. This level helps publishing an artifact provenance in a common format, [in-toto](https://github.com/in-toto/in-toto). For that we need a more explicit `build` and `packaging` steps, which developers can develop further based on their needs. This would satisfy the [Scripted Build](https://slsa.dev/requirements#scripted-build) requirement.

Once the repo meets level 1, we can use [in-toto](https://github.com/in-toto/in-toto) to generate the provenance and verify it.

behnazh:
This PR

- refactors the build and release Actions workflows to remove unnecessary third-party GitHub Actions.
- generates SLSA level 3 provenances.

Closes #25


.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
...
+        WHEEL_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name "*.whl")
+        SBOM_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name "*-sbom.json")
+        HTML_DOCS_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name *-docs-html.zip)
+        DIGEST=$(sha256sum $TARBALL_PATH $WHEEL_PATH $SBOM_PATH $HTML_DOCS_PATH | base64 -w0)
+        echo "digest of artifacts is $DIGEST."
+        echo "::set-output name=artifacts-sha256::$DIGEST"
+    - name: Upload the package artifact
+      # For now only generate artifacts for Python 3.10 on ubuntu-latest.
+      # Currently reusable workflows do not support setting strategy property from the caller workflow.
+      # See: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations.
```
jenstroeger, line 93: Maybe hoist this comment to the top of the file? We’ve mentioned this up there too, I think.

.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
...
+      with:
+        name: artifact-${{ matrix.os }}-${{ matrix.python }}
+        path: ${{ env.GENERATED_PACKAGE_DIR }}
+        if-no-files-found: error
+        retention-days: 1
+    - name: Update release tag
+      if: ${{ github.ref == 'refs/heads/main' }}
+      run: |
+        git push
+        git push --tags
```
jenstroeger, line 105: We’ll add the
```bash
git checkout staging
git rebase main
git push
```
later, in a separate PR?

.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
...
       run: make setup
+    - name: Create changelog and bump
+      if: ${{ github.ref == 'refs/heads/main' }}
+      run: |
+        git config --global user.name "Action User"
+        git config --global user.email "action@github.com"
+        git config --list --global
+        cz changelog
+        cz bump
     - name: Build the package
```
jenstroeger, line 72: Shouldn’t we check, test, build the package _before_ we bump?

.github/workflows/pull-request.yaml
```diff
@@ -33,6 +33,10 @@ jobs:
       with:
         commitlintRulesPath: ./commitlint.rules.js
         commitTitleMatch: 'true'
+
   build:
     needs: pr
     uses: ./.github/workflows/build.yaml
+    permissions:
+      contents: write
```
jenstroeger, line 41: Why is `write` required here?

.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
         PYTHON: python
     - name: Install dependencies
       run: make setup
+    - name: Create changelog and bump
+      if: ${{ github.ref == 'refs/heads/main' }}
```
jenstroeger, line 65: This ensures that the step runs only when the target is the `main` branch?

Should we maybe also consider checking [`github.event.type`](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#event-object-common-properties)?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+      contents: write
+    secrets: inherit
+
+  provenance:
+    needs: build
+    permissions:
+      actions: read
+      id-token: write
+      contents: read
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@d995948e8d53cc639c0d3ef69db31dbc243519c4
```
jenstroeger, line 27: Hmmm… interesting: you can reach directly into another repo’s Action `.yaml` and run it 🤔

They don’t have an actual action we could work with?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+        egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+    - name: Checkout
+      uses: actions/checkout@d0651293c4a5a52e711f25b41b05b2212f385d28
+      with:
+        fetch-depth: 0
+        ref: main
+    - name: Download provenance
+      uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741
+      with:
+        name: ${{needs.provenance.outputs.attestation-name}}
```
jenstroeger, line 51: Spaces inside `${{ }}` (just to be consistent) 😉

.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+        name: ${{ env.ARTIFACT_NAME }}
+        path: ${{ env.GENERATED_PACKAGE_DIR }}
+    - name: Verify the artifact hash
+      env:
+        ARTIFACT_HASH: ${{ needs.build.outputs.artifacts-sha256 }}
+      run: |
+        set -euo pipefail
+        echo "hash of package should be ${{ env.ARTIFACT_HASH }}"
+        # Verify hashes.
+        echo "${{ env.ARTIFACT_HASH }}" | base64 -d | sha256sum --strict --check --status || exit -2
```
jenstroeger, line 64: So here, the `sha256sum` checks the given artifact hashes against the downloaded artifacts themselves, right?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Upload assets
+      env:
+        GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
+      run: |
+        gh release create `git describe --tags --abbrev=0` ./${{ env.GENERATED_PACKAGE_DIR }}/* ${{needs.provenance.outputs.attestation-name}} -F CHANGELOG.md
```
jenstroeger, line 69: Can we please use `--notes-file` instead of `-F`? It’s more descriptive.

It’s not a draft release, but will publish the release? Which in turn will trigger the [`release` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) with `"published"` type?

README.md
```diff
@@ -66,7 +67,7 @@ A _shared package_ or library is intended to be imported by another package or a
...
 
-In the future, the generated `requirements.txt` file with its integrity hash for every dependent package will become an important provenance material to provide transparency in the packaging process (see also [SBOM + SLSA](https://slsa.dev/blog/2022/05/slsa-sbom)).
+The generated `requirements.txt` file with its integrity hash for every dependent package is used to generate SBOM in [CycloneDX format](https://cyclonedx.org/). This is an important provenance material to provide transparency in the packaging process (see also [SBOM + SLSA](https://slsa.dev/blog/2022/05/slsa-sbom)).
```
jenstroeger, line 70: Maybe make `SBOM` a link to https://www.cisa.gov/sbom ?

pyproject.toml
```diff
@@ -41,6 +41,8 @@ dev = [
     "hashin ==0.17.0",
     "mypy >=0.921,<=0.961",
     "pylint >=2.9.3,<=2.14.4",
+    "commitizen >=2.27.1,<3",
+    "cyclonedx-bom>=3.5.0,<4",
```
jenstroeger, line 45: Space 😉

pyproject.toml
```diff
@@ -158,13 +160,15 @@ disable = [
     "too-many-statements",
 ]
 
+
```
jenstroeger, line 163: Remove the empty line because the “pylint” contains multiple sections spaced with a single line. Same below.

jenstroeger (commented):
This is exciting, thank you @behnazh! First read-through comments but I’d like spend some more time on it to understand some of the details better.

One note that I couldn’t add in line: I would add `reopened` here: https://github.com/jenstroeger/python-package-template/blob/1cd7aa0cb52b2395aeef33c8f46b914ac45b92e4/.github/workflows/pull-request.yaml#L10-L13

jenstroeger:
> generates SLSA level 3 provenances.

Also, it looks like we‘re pretty close to [level 4](https://slsa.dev/spec/v0.1/requirements#summary-table) then?

jenstroeger:
Refactoring the [pull-request.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/pull-request.yaml) to remove the dependency on [conventional-pull-request-action](https://github.com/CondeNast/conventional-pull-request-action) is a separate PR, huh?

.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
...
+    - name: Compute package hash
+      if: ${{ matrix.python == '3.10' && matrix.os == 'ubuntu-latest' }}
+      id: compute-hash
+      shell: bash
+      run: |
+        set -euo pipefail
+        TARBALL_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name "*.tar.gz")
+        WHEEL_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name "*.whl")
+        SBOM_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name "*-sbom.json")
+        HTML_DOCS_PATH=$(find ${{ env.GENERATED_PACKAGE_DIR }} -name *-docs-html.zip)
```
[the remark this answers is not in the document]
behnazh: Yes, it should resolve to a single file path.

.github/workflows/build.yaml
behnazh: I'll add it to this PR.

.github/workflows/build.yaml
behnazh: You know the answer ;-)

.github/workflows/release.yaml
behnazh: `sha256sum --strict --check` takes the files, generates hashes, and compares hashes to verify they're the same.

.github/workflows/release.yaml
behnazh: > Can we please use `--notes-file` instead of `-F`? It’s more descriptive.

Yes!

behnazh: > It’s not a draft release, but will publish the release? Which in turn will trigger the [`release` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) with `"published"` type?

Yes, I think so.

behnazh:
> > generates SLSA level 3 provenances.
> 
> Also, it looks like we‘re pretty close to [level 4](https://slsa.dev/spec/v0.1/requirements#summary-table) then?

That needs to be assessed more carefully, but we are close :slightly_smiling_face:

pyproject.toml
```diff
@@ -182,11 +186,13 @@ testpaths = [
...
+# https://github.com/commitizen-tools/commitizen
+# https://commitizen-tools.github.io/commitizen/bump/
+[tool.commitizen]
+bump_message = "bump: release $current_version → $new_version"
+tag_format = "v$major.$minor.$patch$prerelease"
+update_changelog_on_bump = true
+version_files = [
+    "src/package/__init__.py:__version__",
+]
+version = "2.1.0"
```
jenstroeger, line 198: I try to sort the entries in the `pyproject.toml` file alphabetically, so maybe we can move this one after

https://github.com/jenstroeger/python-package-template/blob/1cd7aa0cb52b2395aeef33c8f46b914ac45b92e4/pyproject.toml#L76-L78

.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+      actions: read
+      id-token: write
+      contents: read
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@d995948e8d53cc639c0d3ef69db31dbc243519c4
+    with:
+      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
+
+  release:
+    if: "!startsWith(github.event.commits[0].message, 'bump:')"
+    needs: [build, provenance]
```
jenstroeger, line 33: Isn’t it enough to use
```yaml
needs: provenance
```
because the `provenance` job needs the `build` job to succeed?

.github/workflows/build.yaml
```diff
@@ -46,10 +61,45 @@ jobs:
         PYTHON: python
     - name: Install dependencies
       run: make setup
+    - name: Create changelog and bump
+      if: ${{ github.ref == 'refs/heads/main' }}
+      run: |
+        git config --global user.name "Action User"
+        git config --global user.email "action@github.com"
```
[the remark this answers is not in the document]
jenstroeger: See also question https://github.com/relekang/python-semantic-release/issues/473.

.github/workflows/build.yaml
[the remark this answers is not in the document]
jenstroeger: I guess there’s not a huge performance difference between using [`find`](https://linux.die.net/man/1/find) and [`ls`](https://linux.die.net/man/1/ls):
```
ls ${{ env.GENERATED_PACKAGE_DIR }}/*.tar.gz
```
considering that `env.GENERATED_PACKAGE_DIR` is a directory with only a bunch of files 🤔

.github/workflows/build.yaml
jenstroeger: I do 🤦🏻‍♂️

.github/workflows/pull-request.yaml
behnazh: > 

Because `build.yaml` needs to generate the artifacts and bump the version if it's running on `main`.

.github/workflows/release.yaml
behnazh: > Hmmm… interesting: you can reach directly into another repo’s Action `.yaml` and run it thinking
> 
> They don’t have an actual action we could work with?

Because it's implemented as a reusable workflow I think.

.github/workflows/release.yaml
behnazh: > Isn’t it enough to use
> 
> ```yaml
> needs: provenance
> ```
> 
> because the `provenance` job needs the `build` job to succeed?

No because we need the access the output from both jobs and if we don't add both to `needs`, it will fail.

jenstroeger: Huh 😳 I thought that failure cascades…

.github/workflows/pull-request.yaml
jenstroeger: Oh right, `build.yaml` pushes to the repo 🤦🏻‍♂️

https://github.com/jenstroeger/python-package-template/blob/1cd7aa0cb52b2395aeef33c8f46b914ac45b92e4/.github/workflows/build.yaml#L101-L105

although the [upload-artifact](https://github.com/actions/upload-artifact) Action doesn’t seem to need it?

.github/workflows/build.yaml
jenstroeger: Hmm, so if somebody uses this Action on a repo where the default branch has a different name (e.g. `master`) they’d have to make changes. Hmm.

.github/workflows/pull-request.yaml
behnazh: > although the [upload-artifact](https://github.com/actions/upload-artifact) Action doesn’t seem to need it?

No it doesn't.

.github/workflows/release.yaml
behnazh: > Huh flushed I thought that failure cascades…

The failure cascades, but the outputs are piped only when jobs are specified in `needs` attribute.

jenstroeger: Ah!

behnazh:
@jenstroeger addressed your comments in a88a8f09f0f6bd6f53c164a7d24f54d2dd3cc114.

.github/workflows/build.yaml
```diff
@@ -46,10 +64,55 @@ jobs:
...
       run: make setup
+    - name: Create changelog and bump
+      if: ${{ github.ref == 'refs/heads/main' && matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON }}
+      run: |
+        git config --global user.name "github-actions"
+        git config --global user.email "github-actions@github.com"
+        git config --list --global
+        # First generate the CHANGELOG diff to use as release notes.
+        mkdir -p ${{ env.ARTIFACT_DIR }}
+        cz changelog --dry-run --incremental --unreleased-version "Release Notes" > ${{ env.ARTIFACT_DIR }}/CHANGELOG.md
```
jenstroeger, line 75: The `mkdir` is needed because the `dist/` folder may not exist because `make dist` hasn’t run yet? I wonder if we should add that to our `make setup`?

That generated file `${{ env.ARTIFACT_DIR }}/CHANGELOG.md` isn’t the change log, though, it’s the release notes for this particular run, right?

.github/workflows/build.yaml
```diff
@@ -46,10 +64,55 @@ jobs:
...
+        TARBALL_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*.tar.gz")
+        WHEEL_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*.whl")
+        SBOM_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*-sbom.json")
+        HTML_DOCS_PATH=$(find ${{ env.ARTIFACT_DIR }} -name *-docs-html.zip)
+        CHANGELOG_PATH=$(find ${{ env.ARTIFACT_DIR }} -name CHANGELOG.md)
+        if [ $CHANGELOG_PATH -eq "" ];
+        then
+          touch ${{ env.ARTIFACT_DIR }}/CHANGELOG.md
+          CHANGELOG_PATH=${{ env.ARTIFACT_DIR }}/CHANGELOG.md
+        fi
```
jenstroeger, line 99: If the date of that `CHANGELOG.md` (the release notes, right?) doesn’t matter then you’d safe the check and just run
```
touch ${{ env.ARTIFACT_DIR }}/CHANGELOG.md
CHANGELOG_PATH=$(find ${{ env.ARTIFACT_DIR }} -name CHANGELOG.md)
```

jenstroeger (commented):
Coolio 🤓

.github/workflows/build.yaml
behnazh: > The `mkdir` is needed because the `dist/` folder may not exist because `make dist` hasn’t run yet? I wonder if we should add that to our `make setup`?

Yes, that's correct. I'm happy to add it to `make setup` if you think it's better.

behnazh: > That generated file `${{ env.ARTIFACT_DIR }}/CHANGELOG.md` isn’t the change log, though, it’s the release notes for this particular run, right?

Yes, it's the changelog diff that we use as release notes.

jenstroeger: > I'm happy to add it to `make setup` if you think it's better.

Yes let’s do that. That’s right in line with issue https://github.com/jenstroeger/python-package-template/issues/262 as well 👍🏼

.github/workflows/build.yaml
```diff
@@ -46,10 +64,55 @@ jobs:
...
-      # We don't need to check and test the package separately
-      # because `make dist` runs those targets first and only
-      # builds the package if they succeed.
-      run: make dist
-    - name: Build the docs
-      run: make docs
+      # We don't need to check and test the package separately because `make dist` runs
+      # those targets first and only builds the package if they succeed.
+      run: |
+        make dist requirements sbom
```
jenstroeger, line 83: I think this line makes issue https://github.com/jenstroeger/python-package-template/issues/261 obsolete.

.github/workflows/build.yaml
behnazh: @jenstroeger see commit [3180dca](https://github.com/jenstroeger/python-package-template/pull/259/commits/3180dcab4f114ec74bf96124e00c0b60a8b12898) for the new changes.

.github/workflows/build.yaml
```diff
@@ -46,10 +64,51 @@ jobs:
...
+    - name: Compute package hash
+      if: ${{ matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON }}
+      id: compute-hash
+      shell: bash
+      run: |
+        set -euo pipefail
+        TARBALL_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*.tar.gz")
+        WHEEL_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*.whl")
+        SBOM_PATH=$(find ${{ env.ARTIFACT_DIR }} -name "*-sbom.json")
+        HTML_DOCS_PATH=$(find ${{ env.ARTIFACT_DIR }} -name *-docs-html.zip)
```
jenstroeger, line 92: @behnazh with https://github.com/jenstroeger/python-package-template/pull/263#discussion_r923930333 in mind, couldn’t we use
```
TARBALL_PATH=$(find $ARTIFACT_DIR -name "*.tar.gz")
```
here instead? And anywhere else where we use `${{ env.BLA }}` inside of a shell 🤔

Makefile
```diff
@@ -65,6 +66,7 @@ setup: force-upgrade
 	pre-commit install
 	pre-commit install --hook-type commit-msg
 	pre-commit install --hook-type pre-push
+	mkdir -p dist
```
jenstroeger, line 69: Should we add an explanation here, that the `build.yaml` action expects this folder to exist. I think a std reference link would be https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives?

jenstroeger (commented):
We’re ready to review, huh? 🥳

jenstroeger:
Two more thoughts:

1. We should make name/email changeable e.g. using environment variables so that users can set the user who commits to the repo: https://github.com/jenstroeger/python-package-template/blob/3180dcab4f114ec74bf96124e00c0b60a8b12898/.github/workflows/build.yaml#L70-L71 That’ll be useful when using an organization’s “dev ops” user for whom special branch protection exemptions exist (e.g. no PR required for that user). If not set, we can fall back to the current values. Python Semantic Release has a [similar setting](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#commit-author).
2. Publishing sdist/wheels to a public or self-hosted [PyPI server](https://pypi.org/) is currently missing, we should add that. Python Semantic Release has that [built in](https://python-semantic-release.readthedocs.io/en/latest/#distributing-release-on-pypi-or-custom-repository) and we’ve disabled it: https://github.com/jenstroeger/python-package-template/blob/e5a047f9b0eeef74d06250eb1b63f73c6b63849b/pyproject.toml#L189 We could use the official [twine](https://github.com/pypa/twine/) package or use [flit’s `publish`](https://flit.pypa.io/en/latest/upload.html) command (which, funnily enough, [sends a POST request](https://github.com/pypa/flit/blob/fb949bdc14b47b0ae830015e78d1e37b223328e3/flit/upload.py#L233-L239)).

jenstroeger:
@behnazh within a workflow job, each step runs as a separate process and shell, yes?

Because if we have one step that creates _and activates_ a virtual environment:

https://github.com/jenstroeger/python-package-template/blob/e5a047f9b0eeef74d06250eb1b63f73c6b63849b/.github/workflows/build.yaml#L33-L39

then the [`.` (source) shell builtin](https://www.gnu.org/software/bash/manual/bash.html#Bourne-Shell-Builtins-1) does the following:

> Read and execute commands from the filename argument in the current shell context.

Note the _in the current shell context_!

I think that shell context closes when the step finishes, and that means the next step 

https://github.com/jenstroeger/python-package-template/blob/e5a047f9b0eeef74d06250eb1b63f73c6b63849b/.github/workflows/build.yaml#L47-L48

does _not_ run inside of an activated venv ([log](https://github.com/jenstroeger/python-package-template/runs/7419949560?check_suite_focus=true#step:8:9)):
```
Makefile:26: No Python virtual environment found, proceeding anyway
```
and so on. Take a look at the remaining steps in any of the recent Action runs [here](https://github.com/jenstroeger/python-package-template/runs/7419949560?check_suite_focus=true#step:9:9) and [here](https://github.com/jenstroeger/python-package-template/runs/7419949560?check_suite_focus=true#step:10:9) 🤦🏻‍♂️

The fix would be to activate the virtual environment in every step, i.e.
```yaml
    - name: Install dependencies
      run: |
        source .venv/bin/activate
        make setup
```
which sucks if we want to continue supporting Windows, or running all remaining build goals in the same step:
```yaml
    - name: Create virtual environment and build artifacts
      if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest' }}
      run: |
        make venv
        source .venv/bin/activate
        make setup
        make dist docs
      env:
        PYTHON: python
```
Asked for confirmation in https://github.com/actions/setup-python/issues/359#issuecomment-1191450368.

.github/workflows/build.yaml
behnazh: Thanks. Yes the env variables in bash don't need the `env` [context](https://docs.github.com/en/actions/learn-github-actions/environment-variables#using-contexts-to-access-environment-variable-values). However, we should be careful about Windows OS because the syntax for PowerShell is `$env:NAME`.

behnazh:
> 1. We should make name/email changeable e.g. using environment variables so that users can set the user who commits to the repo:

:+1:

behnazh:
> 2\. We could use the official [twine](https://github.com/pypa/twine/) package or use [flit’s `publish`](https://flit.pypa.io/en/latest/upload.html) command (which, funnily enough, [sends a POST request](https://github.com/pypa/flit/blob/fb949bdc14b47b0ae830015e78d1e37b223328e3/flit/upload.py#L233-L239)).

I would probably do the POST request because we need to upload the provenance too.

behnazh:
> The fix would be to activate the virtual environment in every step, i.e.

Remind me why we need to activate the venv in Actions workflows?

jenstroeger:
> I would probably do the POST request because we need to upload the provenance too.

…to the PyPI server, in addition to the python package?

> Remind me why we need to activate the venv in Actions workflows?

I think, uhm, because 

- then we’re consistent with the local development and use of the Makefile, and
- we switched to using venvs in the Actions with PR https://github.com/jenstroeger/python-package-template/pull/248 (see the first commit to that PR which breaks, and the second commit which fixed it) because we need the `.venv/` folder to store the `.upgraded-on` file.

behnazh:
> …to the PyPI server, in addition to the python package?

Yes, if we can't accompany our package with the provenance, then how should people find it?

behnazh:
> > 1. We should make name/email changeable e.g. using environment variables so that users can set the user who commits to the repo:
> 
> +1

I addressed you comment [here](https://github.com/jenstroeger/python-package-template/compare/3180dcab4f114ec74bf96124e00c0b60a8b12898..b0bae805034b104645bb7d2125574e71cf2877b6).

Makefile
behnazh: > Should we add an explanation here, that the `build.yaml` action expects this folder to exist. I think a std reference link would be https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives?

Done [here](https://github.com/jenstroeger/python-package-template/compare/3180dcab4f114ec74bf96124e00c0b60a8b12898..b0bae805034b104645bb7d2125574e71cf2877b6).

.github/workflows/build.yaml
jenstroeger: > However, we should be careful about Windows OS because the syntax for PowerShell is `$env:NAME`.

Hmm, I wonder if we can set the shell for steps across platforms, e.g. using [`defaults.run`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun)—would that allow us to use bash on Windows runners as well?

behnazh:
> Does PyPI support the provenance? The [`twine upload` command](https://twine.readthedocs.io/en/latest/#twine-upload) doesn’t seem to and only offers a `--cert` command-line option

Doesn't look like. And `twine` only uploads the wheel and source tarball (and `.asc` files to resign). It throws `InvalidDistribution` for other file formats.

jenstroeger:
> Doesn't look like. And `twine` only uploads the wheel and source tarball (and `.asc` files to resign). It throws `InvalidDistribution` for other file formats.

One more item for our final writeup. It looks like the PyPI infrastructure itself isn’t (yet) equipped for a more thorough provenance management 🤔

behnazh:
Regarding publishing to the PyPI server, I prefer to use `twine upload` because I can explicitly pass the verified artifacts as args and it doesn't try to build the package. However, `flit publish` [might build](https://flit.pypa.io/en/latest/cmdline.html#flit-publish) the package and publish an unverified artifact at the release step.

@jenstroeger WDUT?

jenstroeger:
@behnazh thanks for pointing that out! Interesting (somewhat disagreeable) design choice by `flit`:

> Build a wheel and an sdist (tarball) from the package, and upload them to PyPI or another repository.

I would have expected that `flit publish` fails if there are no artifacts to publish, instead of building them without me asking it. But I can also see the point: I guess in general that’s more convenient to the user? Wonder what @takluyver’s reasoning was.

> I prefer to use `twine upload` because I can explicitly pass the verified artifacts as args and it doesn't try to build the package.

Agreed, let’s do that then.

takluyver:
> I would have expected that flit publish fails if there are no artifacts to publish, instead of building them without me asking it. But I can also see the point: I guess in general that’s more convenient to the user? Wonder what @takluyver’s reasoning was.

Roughly that I want to be thinking about publishing the project, not about creating and uploading artifacts.

To look at it another way, there are two kinds of issue that can arise: one where the artifacts you build and examine are not exactly the same as the artifacts you build and upload, and one where you forget to rebuild artifacts after a last minute change before uploading, or select the wrong artifacts to upload. Flit's approach minimises the risk of the latter, by building artifacts from your source code as part of the publication step. For the former, the aim is to make building artifacts as deterministic as possible, so rebuilding with no changes in the source files should ideally generate identical artifacts.

Finally, if someone really feels the need to check the exact artifacts that will be uploaded, it's easy to use `flit build` to create them and then `twine` to upload them.

.github/workflows/build.yaml
```diff
@@ -18,38 +34,86 @@ jobs:
...
-      # We don't need to check and test the package separately
-      # because `make dist` runs those targets first and only
-      # builds the package if they succeed.
-      run: make dist
-    - name: Build the docs
-      run: make docs
+      # We don't need to check and test the package separately because `make dist` runs
+      # those targets first and only builds the package if they succeed.
+      run: |
+        make dist requirements sbom
```
jenstroeger, line 84: Add `docs` as part of the build?

.github/workflows/build.yaml
```diff
@@ -18,38 +34,86 @@ jobs:
...
+        TARBALL_PATH=$(find dist -name "*.tar.gz")
+        WHEEL_PATH=$(find dist -name "*.whl")
+        SBOM_PATH=$(find dist -name "*-sbom.json")
+        HTML_DOCS_PATH=$(find dist -name *-docs-html.zip)
+        # Make sure dist/RELEASE_NOTES.md (which contains the release notes) exists.
+        touch dist/RELEASE_NOTES.md
+        NOTES_PATH=$(find dist -name RELEASE_NOTES.md)
+        # Compute the sha digest for all the release files and encode them using base64.
+        DIGEST=$(sha256sum $TARBALL_PATH $WHEEL_PATH $SBOM_PATH $HTML_DOCS_PATH $NOTES_PATH | base64 -w0)
+        echo "digest of artifacts is $DIGEST."
```
jenstroeger, line 101: Should the `echo` be in [sentence case](https://en.wikipedia.org/wiki/Capitalization#Sentence_case)? Same question for the `echo` in the Release Action.

.github/workflows/build.yaml
```diff
@@ -18,38 +34,86 @@ jobs:
...
+        name: artifact-${{ matrix.os }}-${{ matrix.python }}
+        path: dist
+        if-no-files-found: error
+        retention-days: 1
+    - name: Update release tag
+      # Push the release tag only once when branch is main on the specified OS and Python version.
+      if: github.ref == 'refs/heads/main' && matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON
+      run: |
+        git push
+        git push --tags
```
jenstroeger, line 119: Do we want to change `staging` in a separate PR?
```bash
git checkout staging
git rebase main
git push
```
This would hardwire the branch name `staging` into the process, though, and maybe we should make it configurable? Probably separate discussion and PR 🤓

.github/workflows/release.yaml
```diff
@@ -0,0 +1,91 @@
...
+  build:
+    uses: ./.github/workflows/build.yaml
+    permissions:
+      contents: write # To create the release tag.
+    secrets: inherit
+
+  provenance:
+    needs: build
+    permissions:
+      actions: read   # To read the workflow path.
```
jenstroeger, line 23: One space for consistency or three spaces for vertical alignment? 😉

.github/workflows/release.yaml
```diff
@@ -0,0 +1,91 @@
...
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    # The generator should be referenced with a semantic version.
+    # The build will fail if we reference it using the commit sha.
```
jenstroeger, line 27: Dang! Is that a Github problem?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,91 @@
...
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    with:
+      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
+
+  release:
+    if: "!startsWith(github.event.commits[0].message, 'bump:')"
```
jenstroeger, line 33: Do we need the `"` around the expression here? It’s JS, isn’t it?

README.md
```diff
@@ -210,29 +211,40 @@ open docs/_build/html/index.html
...
 
-This builds a source package and a binary distribution, and stores the files in your local `dist/` folder.
+## Build integrity using SLSA framework
+
+The build process in this repository follows the requirements in the [SLSA framework](https://slsa.dev/) to be compliant at level 3. An important aspect of SLSA to improve the supply chain security posture is to generate a verifiable provenance for the build pipeline. Such a provenance can be used to verify the builder and let the consumers check the materials and configurations used while building an artifact. In this repository we use the [generic provenance generator reusable workflow](https://github.com/slsa-framework/slsa-github-generator) to generate a provenance that can attest to the following artifacts in every release:
+
+* Binary dist (wheel)
+* Source dist (tarball)
+* SBOM (CycloneDx format)
+* HTML Docs
```
jenstroeger, line 240: In other parts of the Markdown we use `-` instead of `*` for list items. OCD! OCD!

README.md
```diff
@@ -210,29 +211,40 @@ open docs/_build/html/index.html
...
+
+To verify the artifact using the provenance follow the instructions in the [SLSA verifier](https://github.com/slsa-framework/slsa-verifier) project to install the verifier tool. After downloading the artifacts and provenance, verify each artifact individually, e.g.,:
+
+```bash
+slsa-verifier -artifact-path  ~/Downloads/package-2.2.0.tar.gz -provenance attestation.intoto.jsonl -source github.com/jenstroeger/python-package-template
```
jenstroeger, line 245: Maybe remove the `~/Downloads/` because that’s platform dependent?

jenstroeger (commented):
This is an exciting change, thank you for all your work @behnazh, and thanks for adding version comments after Action SHAs.

One point of discussion maybe: how do we want to call the Action files — by what they do (check and release) or by what triggers them (pull request) 🤔

We have this comment still open: https://github.com/jenstroeger/python-package-template/pull/259#discussion_r922649527

In the [build.yaml](https://github.com/jenstroeger/python-package-template/blob/generate-provenance/.github/workflows/build.yaml) we install the venv dependent on the Runner (Windows vs Mac vs Linux) and then activate the venv

https://github.com/jenstroeger/python-package-template/blob/70ccc0208350abc442c5b0a524f9c0c4996e0556/.github/workflows/build.yaml#L52-L65

That doesn’t work though because the `run:` shell is closed and the next step uses its own new shell which doesn’t have an activated venv (see also https://github.com/jenstroeger/python-package-template/pull/259#issuecomment-1191455334 ). I think we have two options:

1. Activate the venv at the beginning of every step; or
2. Don’t create the venv and simply operate in the Action container’s global Python installation

We did that because https://github.com/jenstroeger/python-package-template/pull/259#issuecomment-1191990312. What now?

.github/workflows/build.yaml
behnazh: Yes, that should be done in a separate PR.

.github/workflows/release.yaml
behnazh: The reason why this expression needs `"` is the `!` which is a reserved symbol in YAML.

README.md
behnazh: That's just an example to show they have to download it.

.github/workflows/release.yaml
behnazh: > 2\. Don’t create the venv and simply operate in the Action container’s global Python installation

I'll go with this option.

behnazh: > 

The `Makefile` expects `.venv/upgraded-on`, so we need to investigate this in a separate PR. I left a comment in `build.yaml`.

behnazh:
@jenstroeger I addressed your comments. Please see 61002f72a7ff68cb8138b509b02de1e1efcf8c61.

.github/workflows/build.yaml
jenstroeger: As per conversation, let’s leave a note in the README that this template repository assumes `main` to be the default branch.

.github/workflows/build.yaml
jenstroeger: Issue https://github.com/jenstroeger/python-package-template/issues/266.

.github/workflows/release.yaml
jenstroeger: Issue https://github.com/jenstroeger/python-package-template/issues/267.

jenstroeger (approved):

PR #263, opened 2022-07-18
refactor(ci): use commitizen tool for pull request action

jenstroeger:
Remove the [conventional-pull-request-action](https://github.com/CondeNast/conventional-pull-request-action) from our Pull Request Action, and instead use [commitizen](https://github.com/commitizen-tools/commitizen) (which we already use from our git hooks).


jenstroeger (commented):
I pinned `commitizen` as part of the `"dev"` optional dependencies for now, because our `Makefile` doesn’t differentiate between different setups anyway.

Also, I’ve not yet noodled through the [Settings](https://commitizen-tools.github.io/commitizen/config/#settings) yet to check if we can/should configure this Action run…

.github/workflows/pull-request.yaml
```diff
@@ -1,38 +1,47 @@
...
 
 name: Pull Request
 on:
   pull_request:
     branches: [main, staging]
     types:
     - opened
+    - reopened
     - edited
     - synchronize
```
jenstroeger, line 11: I also noticed that we’re using two different notations for [YAML block collections](https://yaml.org/spec/1.1/current.html#id857181) (in this case sequences):
```yaml
branches: [main, staging]
```
is what the spec calls a [“flow style”](https://yaml.org/spec/1.1/current.html#flow%20style/information%20model) equivalent to writing out the sequence itself in “block style”:
```yaml
types:
- opened
- reopened
- edited
- synchronize
```
Should we agree on one consistent notation?

.github/workflows/pull-request.yaml
```diff
@@ -1,38 +1,47 @@
...
+    - name: Create and activate virtual environment
+      run: |
+        make venv
+        source .venv/bin/activate
+      env:
+        PYTHON: python
+    - name: Install dependencies
+      run: make setup
+    - name: Check PR title
+      run: echo "${{ github.event.pull_request.title }}" | cz check
```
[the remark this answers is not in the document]
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/263/commits/51da860a2bd61ed90cece33b0f712f5efb3a1dae.

.github/workflows/pull-request.yaml
behnazh: > Should we agree on one consistent notation?

Sure, I'm happy with either.

.github/workflows/pull-request.yaml
```diff
@@ -1,38 +1,47 @@
...
-        commitTitleMatch: 'true'
+        python-version: '3.10'
+    - name: Create and activate virtual environment
+      run: |
+        make venv
+        source .venv/bin/activate
+      env:
+        PYTHON: python
+    - name: Install dependencies
+      run: make setup
```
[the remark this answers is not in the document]
behnazh: I also think it's useless to install the whole dependencies here. We could either install `commitizen` or add a new make target for conventional commits.

We should also consider if `harden-runner` is adding any value here. Since we don't use an external Actions workflow, I think we can disable it? We could do the same on other places where we don't use a third-party workflow?

.github/workflows/pull-request.yaml
```diff
@@ -1,38 +1,52 @@
...
+    - name: Check PR title
+      run: echo "$PR_TITLE" | cz check
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
+    - name: Check PR commit messages
+      run: cz check --rev-range origin/$PR_BASE_BRANCH_NAME..origin/$PR_HEAD_BRANCH_NAME
+      env:
+        PR_BASE_BRANCH_NAME: ${{ github.event.pull_request.base.ref }}
```
behnazh, line 47: Can we inline these variables?

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dev = [
     "hashin ==0.17.0",
     "mypy >=0.921,<=0.961",
     "pylint >=2.9.3,<=2.14.4",
+    "commitizen ==2.28.0",
```
behnazh, line 44: Why not `"commitizen >=2.28.0,<3"`?

.github/workflows/pull-request.yaml
jenstroeger: I think inlining would mean to open the door to injection, see also the [**Remidiation** section in this blog](https://securitylab.github.com/research/github-actions-untrusted-input/#remediation). The initial code of this PR has these values inlined:

https://github.com/jenstroeger/python-package-template/blob/2f357e1465d2a7f252304788141b5549c41cc36a/.github/workflows/pull-request.yaml#L40-L43

With a PR title like 
```
chore: `echo test`
```
we’d have Github expand the `run:` to the (harmless but injected) shell command
```
echo chore: `echo test` | cz check 
```
That’s what we want to avoid.

.github/workflows/pull-request.yaml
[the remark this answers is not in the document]
jenstroeger: Commits https://github.com/jenstroeger/python-package-template/pull/263/commits/3d58166eb30baf037498bb95f7024ac342e0bb09 and https://github.com/jenstroeger/python-package-template/pull/263/commits/40890a364077b409fa3176447731772daccde871.

.github/workflows/pull-request.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/263/commits/a6103832b49ab416cfa51398fa68d22243fd1d90: for now I switched to “block style”.

pyproject.toml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/263/commits/40890a364077b409fa3176447731772daccde871 pins commitizen to v2.28.0 inside the Action itself. For consistency and because we run commitizen as a standalone app, I pinned it to a version instead of a range.

.github/workflows/pull-request.yaml
jenstroeger: See also https://github.com/jenstroeger/python-package-template/pull/263#discussion_r923930333 above.

.github/workflows/pull-request.yaml
```diff
@@ -1,38 +1,46 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08 # v4.1.0
       with:
-        commitlintRulesPath: ./commitlint.rules.js
-        commitTitleMatch: 'true'
+        python-version: '3.10'
+    - name: Set up Commitizen
+      run: |
+        pip install --upgrade pip wheel
+        pip install 'commitizen ==2.28.0'
```
behnazh, line 33: Can you add comments to explain how these steps work?

jenstroeger: Like this: commit https://github.com/jenstroeger/python-package-template/pull/263/commits/aa5b0eebd273af52da61ac9c5e08da986cd0a1ee?

behnazh: I meant to explain why you do `pip install 'commitizen ==2.28.0'` directly for example instead of using `make setup`.

pyproject.toml
behnazh: As discussed, please turn that to a comment in code.

.github/workflows/pull-request.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/263/commits/0a1d3cdf48b3e7b18360d6d3264a4ef6759d54c8.

pyproject.toml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/263/commits/0a1d3cdf48b3e7b18360d6d3264a4ef6759d54c8.

behnazh (approved):

PR #269, opened 2022-08-03
fix: enable CI on release bump commit

behnazh:


.github/workflows/build.yaml
```diff
@@ -71,7 +76,7 @@ jobs:
     - name: Install dependencies
       run: make setup
     - name: Create changelog and bump
-      if: github.ref == 'refs/heads/main' && matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.python == '3.10' && !startsWith(github.event.commits[0].message, 'bump:') }}
```
jenstroeger, line 79: Dang! I thought that `if:` interprets its value as JS?

jenstroeger (approved):

jenstroeger:
Also, I think we should turn

https://github.com/jenstroeger/python-package-template/blob/c7541754294c634cd807068ffaf37bd0279cbb52/.github/workflows/build.yaml#L84

into
```
cz bump --yes 
```
because on the very first bump no tags exist yet and the Action would fail:
```
> cz bump
Tag v0.0.0 could not be found. 
Possible causes:
- version in configuration is not the current version
- tag_format is missing, check them using 'git tag --list'

? Is this the first tag created? (Y/n)                                                                               
```

.github/workflows/build.yaml
behnazh: > Dang! I thought that `if:` interprets its value as JS?

It does, but notice the `!` which is a reserved symbol in YAML, hence using `${{}}`.

behnazh: > And you wired the OS and Python, instead of using the variables — only here, and not in other lines?

:facepalm:

jenstroeger: Why do I keep asking the same question 🤦🏻‍♂️

jenstroeger:
And as a follow-up, check this out issue https://github.com/commitizen-tools/commitizen/issues/549 😳

jenstroeger (approved):

PR #271, opened 2022-08-04
ci: enable checks on pushes to staging branch

behnazh:


jenstroeger (approved):
Looks good. Should we re-rename the action to `check-and-release.yaml` again because it’s not only used for “releasing” anymore?

PR #276, opened 2022-08-11
ci: build requirements.txt and SBOM artifacts only on release

jenstroeger:
It’s probably enough to build the `requirements.txt` file and the SBOM only when we create a release…

@behnazh we currently build and upload the release artifacts with _every run_ of the workflow, and only the “Update release tag” step is guarded

https://github.com/jenstroeger/python-package-template/blob/c7541754294c634cd807068ffaf37bd0279cbb52/.github/workflows/build.yaml#L124

Should we limit the other steps too?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -92,15 +92,16 @@ jobs:
       # those targets first and only builds the package if they succeed.
       run: |
         # Build the sdist and wheel distribution of the package and docs as a zip file.
-        # Generate the requirements.txt that contains the hash digests of the dependencies.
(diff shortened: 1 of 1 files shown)


jenstroeger:
Hmm… maybe another reason why we want to build the `requirements.txt` file only on release, i.e. only on Linux: https://github.com/jenstroeger/python-package-template/runs/7787360179?check_suite_focus=true#step:9:166

behnazh (approved):
We need to be more careful with the `refactor` scope in the commit message as it ends up in the CHANGELOG.md by default. 

Wouldn't `ci` bee enough here?

PR #283, opened 2022-08-17
chore: add reviewers to the Dependabot configuration

jenstroeger:
We can assign one or more reviewers to PRs that Dependabot opens ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers)), and perhaps it’s worthwhile adding to the configuration? Not sure if we want to comment that out or leave it as an active feature.

We we add this change, we probably should document it in our README.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -14,6 +14,9 @@ updates:
     include: scope
   open-pull-requests-limit: 13
   target-branch: staging
+  reviewers:
+  - jenstroeger
+  - behnazh
 
 - package-ecosystem: github-actions
   directory: /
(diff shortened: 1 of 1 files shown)


behnazh (commented):
I'm not sure adding our usernames to a template repo that is meant to be used by others is a good idea :neutral_face:

jenstroeger:
Or comment it out. I just wanted this in there for documentation purposes (and because I know we’ll use it in other repos).

behnazh:
> Or comment it out. I just wanted this in there for documentation purposes (and because I know we’ll use it in other repos).

SGTM.

behnazh (approved):

PR #285, opened 2022-08-18
chore: use actual .venv directory name in nuke-caches Makefile recipe

jenstroeger:
This is to be consistent with the remainder of the Makefile where we’ve established the convention that the Python virtual environment is the `.venv/` folder. The `VIRTUAL_ENV` environment variable, in contrast, is used only to check whether a virtual env was _activated_ but not to determine its location.


jenstroeger:
@behnazh nice to see that there are no more complaints if the PR title differs from the commit title 😉

behnazh (approved):

PR #291, opened 2022-08-23
refactor(ci): remove write permissions from build.yaml

behnazh:


.github/workflows/release.yaml
```diff
@@ -11,13 +11,71 @@ permissions:
...
+    - name: Upload the Release Notes
+      uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+      with:
+        name: ${{ env.ARTIFACT_NAME }}
+        path: dist
+        if-no-files-found: error
+        retention-days: 1
+
+
+
```
jenstroeger, line 71: Maybe just one line?

.github/workflows/release.yaml
```diff
@@ -32,7 +90,6 @@ jobs:
       base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
 
   release:
-    if: ${{ github.ref == 'refs/heads/main' && !startsWith(github.event.commits[0].message, 'bump:') }}
```
jenstroeger, line 35: Removing this guard is ok because `build` may not run, therefore `release` doesn’t run?

jenstroeger (commented):
Looks good 🥳

.github/workflows/release.yaml
behnazh: > Removing this guard is ok because `build` may not run, therefore `release` doesn’t run?

Yes, that's correct. Here is the dependency chain which saves us from repeating the guard:
```
check -> bump (guarded) -> build -> provenance -> release
```

jenstroeger:
This change won’t address the [failing Dependabot action runs](https://github.com/jenstroeger/python-package-template/runs/7947365203?check_suite_focus=true), yes?

behnazh:
> This change won’t address the [failing Dependabot action runs](https://github.com/jenstroeger/python-package-template/runs/7947365203?check_suite_focus=true), yes?

This could be fixed now as we don't need the secrets with write permissions anymore. From this [blog](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/).

>workflow runs that are triggered by Dependabot from push, pull_request, pull_request_review, or pull_request_review_comment events will be treated as if they were opened from a repository fork. This means they will receive a read-only GITHUB_TOKEN and will not have access to any secrets available in the repository. This will cause any workflows that attempt to write to the repository to fail.

But I'm not sure, and we need to try it out.

jenstroeger:
Also, should we change these 

https://github.com/jenstroeger/python-package-template/blob/9b08ae2dc65793b2c9232a630af274bff6f0f63d/.github/workflows/build.yaml#L68-L69

https://github.com/jenstroeger/python-package-template/blob/9b08ae2dc65793b2c9232a630af274bff6f0f63d/.github/workflows/build.yaml#L75-L77

to single-lines:
```yaml
run: <cmd>
```
instead of using a [Literal Block Scalar](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics) that just contains a single command?

.github/workflows/release.yaml
```diff
@@ -11,13 +11,71 @@ permissions:
...
+        pip install --upgrade pip wheel
+        pip install 'commitizen ==2.32.1'
+    - name: Create changelog and bump
+      run: |
+        # Set up the GitHub user and email as author for the release commit.
+        git config --global user.name $USER_NAME
+        git config --global user.email $USER_EMAIL
+        git config --list --global
+        # First generate the CHANGELOG diff to use as release notes.
+        mkdir -p dist
```
jenstroeger, line 54: Technically we don’t need `dist/` here anymore, do we?

behnazh: We actually need it because we don't call `make setup` here for efficiency concerns.

jenstroeger:
Hmm, I think we could remove `commitizen` and `twine` from the dependencies in `pyproject.toml`, they’re installed and used _only_ in Release and PR Actions 🤔

behnazh:
> Hmm, I think we could remove `commitizen` and `twine` from the dependencies in `pyproject.toml`, they’re installed and used _only_ in Release and PR Actions thinking

I think they need to be there, so that they can get recorded in the (s/i)bom though. They're used for creating the release commit and uploading the artifacts.

jenstroeger:
> I think they need to be there, so that they can get recorded in the (s/i)bom though. They're used for creating the release commit and uploading the artifacts.

It would probably make sense to document that, e.g. a note in the `pyproject.toml` to remind users to that the Actions install their own versions? We want to make sure that the BOM is consistent.

behnazh:
> > I think they need to be there, so that they can get recorded in the (s/i)bom though. They're used for creating the release commit and uploading the artifacts.
> 
> It would probably make sense to document that, e.g. a note in the `pyproject.toml` to remind users to that the Actions install their own versions? We want to make sure that the BOM is consistent.

Agreed. Would it be better to be addressed in the other [PR](https://github.com/jenstroeger/python-package-template/pull/284)?

jenstroeger (approved):

PR #314, opened 2022-09-07
fix: remove requirements.txt when cleaning the distribution artifacts

jenstroeger:
We use the `requirements.txt` file as a distribution artifact, so we should delete it as part of the distribution too. Methinks.


behnazh (approved):
Maybe rephrase the commit message? It's not very clear what you mean by distribution.

PR #322, opened 2022-09-13
chore: use consistent notation to run commands

jenstroeger:


behnazh (approved):

jenstroeger:
@behnazh apparently [`shellcheck`](https://www.shellcheck.net/) suggests otherwise: [Use $(...) notation instead of legacy backticked \`...\`](https://www.shellcheck.net/wiki/SC2006). I’m tempted to change?

PR #323, opened 2022-09-13
chore(ci): add Harden Runner when running the PR Action workflow

jenstroeger:
I don’t remember: did we intentionally _not_ call the [Harden Runner](https://github.com/step-security/harden-runner) Action here?

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -22,6 +22,10 @@ jobs:
     name: Check PR title and commit messages
     runs-on: ubuntu-latest
     steps:
+    - name: Harden Runner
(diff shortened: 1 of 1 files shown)


behnazh:
> I don’t remember: did we intentionally _not_ call the [Harden Runner](https://github.com/step-security/harden-runner) Action here?

I think we removed it because we use commitizen Python package instead of the external GHA PR title linter.

jenstroeger:
So we trust `cz` to behave, and don’t need to monitor/harden that particular runner?

Close this issue, maybe leave a comment in the Action?

behnazh (approved):

jenstroeger:
Closing this without merging as per above conversation.

PR #324, opened 2022-09-13
chore(ci): reformat and split up steps to improve debugging & readability

jenstroeger:
One more thought: it looks like Github [is not able to share environment variables _between_ workflows](https://docs.github.com/en/actions/learn-github-actions/environment-variables). But I wonder if perhaps we should move “global” variables like this:

https://github.com/jenstroeger/python-package-template/blob/ceee4a7b803747f9e168f188227345685d984c5b/.github/workflows/release.yaml#L14-L17

to the only step that actually uses them

https://github.com/jenstroeger/python-package-template/blob/ceee4a7b803747f9e168f188227345685d984c5b/.github/workflows/release.yaml#L57-L61

Simply for the sake of tightening the scope.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,11 +1,23 @@
 # This is a trusted builder implemented as a reusable workflow that can be called by other
 # Actions workflows. It checks, tests, and builds the artifacts including SBOM and documentations,
-# and computes hash digests as output to be used by a SLSA provenance generator. The artifacts are always uploaded
-# for every job to be used for debugging purposes, but they will be removed within the specified retention days.
-# Even though we run the build in a matrix to check against different platforms, due to a known limitation of
-# reusable workflows that do not support setting strategy property from the caller workflow, we
-# only generate artifacts for ubuntu-latest and Python 3.10, which can be used to create a release.
-# See: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations.
+# and computes hash digests as output to be used by a SLSA provenance generator. The artifacts are
+# always uploaded for every job to be used for debugging purposes, but they will be removed within
+# the specified retention days.
+#
+# Even though we run the build in a matrix to check against different platforms, due to a known
+# limitation of reusable workflows that do not support setting strategy property from the caller
+# workflow, we only generate artifacts for ubuntu-latest and Python 3.10, which can be used to
+# create a release. For details see:
+#
+# https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations
+#
+# Note: if the build workflow needs to access secrets, they need to be passed by the caller using
+# `secrets: inherit`. See also
+#
+# https://docs.github.com/en/actions/using-workflows/reusing-workflows
+# https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
+#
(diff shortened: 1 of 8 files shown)


jenstroeger:
Oh, and should we perhaps split up [release.yaml](https://github.com/jenstroeger/python-package-template/blob/reformat-action-steps/.github/workflows/release.yaml) into two — one for “push to `staging`“ (which isn’t really a release) and one for “push to `main`” (which is the actual release)?

.github/workflows/build.yaml
```diff
@@ -55,39 +71,50 @@ jobs:
...
-        # Find the paths to the files that will be included in the release.
         TARBALL_PATH=$(find dist -name "*.tar.gz")
         WHEEL_PATH=$(find dist -name "*.whl")
         REQUIREMENTS_PATH=$(find dist -name "*-requirements.txt")
         SBOM_PATH=$(find dist -name "*-sbom.json")
-        HTML_DOCS_PATH=$(find dist -name *-docs-html.zip)
-        BUILD_EPOCH_PATH=$(find dist -name *-build-epoch.txt)
-        # Compute the sha digest for all the release files and encode them using base64.
+        HTML_DOCS_PATH=$(find dist -name "*-docs-html.zip")
+        BUILD_EPOCH_PATH=$(find dist -name "*-build-epoch.txt")
```
jenstroeger, line 113: Consider adding `-type f` as cmd line argument for `find` to constrain the search.

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/324/commits/5f462d5b97a64f9e7a1292fcc5ff08979bbd8110.

behnazh:
> One more thought: it looks like Github [is not able to share environment variables _between_ workflows](https://docs.github.com/en/actions/learn-github-actions/environment-variables). But I wonder if perhaps we should move “global” variables like this:
> 
> https://github.com/jenstroeger/python-package-template/blob/ceee4a7b803747f9e168f188227345685d984c5b/.github/workflows/release.yaml#L14-L17
> 
> to the only step that actually uses them
> 
> https://github.com/jenstroeger/python-package-template/blob/ceee4a7b803747f9e168f188227345685d984c5b/.github/workflows/release.yaml#L57-L61
> 
> Simply for the sake of tightening the scope.

I think the reason we had put it at the top was to make it easy for people to see the inputs they need to change in the workflow.

behnazh:
> Oh, and should we perhaps split up [release.yaml](https://github.com/jenstroeger/python-package-template/blob/reformat-action-steps/.github/workflows/release.yaml) into two — one for “push to `staging`“ (which isn’t really a release) and one for “push to `main`” (which is the actual release)?

That would result in some duplication but if that makes things simpler, I'm happy to do it.

behnazh (approved):

jenstroeger:
> > Oh, and should we perhaps split up [release.yaml](https://github.com/jenstroeger/python-package-template/blob/reformat-action-steps/.github/workflows/release.yaml) into two — one for “push to `staging`“ (which isn’t really a release) and one for “push to `main`” (which is the actual release)?
> 
> That would result in some duplication but if that makes things simpler, I'm happy to do it.

We can discuss in a separate issue if the need arises…

PR #329, opened 2022-09-20
fix: a Makefile’s SHELL variable is not an executable shebang

jenstroeger:
I stumbled upon issues when I played with the [`.ONESHELL:` special target](https://www.gnu.org/software/make/manual/html_node/One-Shell.html):
```
> make audit
make: /usr/bin/env bash: No such file or directory
...
```
Documentation on this is a little sparse, I thought.

<hr>

I also think we should consider using the special target `.ONESHELL:`: it runs the entire recipe in a _single_ shell which would allow us to set & use environment variables within that recipe. So far we’ve danced around `make` running every recipe line in its own shell, and that has complicated some of our process. In that context, it would also make sense to discuss [`.EXPORT_ALL_VARIABLES:`](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html).

There’s also an option to [silence `make`’s command-line echoing](https://www.gnu.org/software/make/manual/html_node/Echoing.html), which may be usful?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 
 # Use bash as the shell when executing a rule's recipe. For more details:
 # https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
-# https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang
-SHELL := /usr/bin/env bash
+SHELL := bash
+.SHELLFLAGS += -e
 
 # Set the package's name and version for use throughout the Makefile.
 PACKAGE_NAME=package
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 # Use bash as the shell when executing a rule's recipe. For more details:
 # https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
 SHELL := bash
-.SHELLFLAGS += -e
 
 # Set the package's name and version for use throughout the Makefile.
 PACKAGE_NAME=package


Makefile
```diff
@@ -1,8 +1,8 @@
 
 # Use bash as the shell when executing a rule's recipe. For more details:
 # https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
-# https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang
-SHELL := /usr/bin/env bash
+SHELL := bash
+.SHELLFLAGS += -e
```
behnazh, line 5: Can you please add a comment?

behnazh:
> In that context, it would also make sense to discuss [`.EXPORT_ALL_VARIABLES:`](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html).

Not sure what would be the purpose?

Makefile
jenstroeger: Let me remove this flag for now. From the [docs](https://www.gnu.org/software/make/manual/html_node/One-Shell.html):

> You can modify `.SHELLFLAGS` to add the `-e` option to the shell which will cause any failure anywhere in the command line to cause the shell to fail, but this could itself cause your recipe to behave differently. Ultimately you may need to harden your recipe lines to allow them to work with `.ONESHELL`.

So perhaps we should have a separte discussion about `.ONESHELL` and `.SHELLFLAGS` and exporting variables to the subshells.

Commit https://github.com/jenstroeger/python-package-template/pull/329/commits/1ebcf37b1cb24de5ecc7c210389b8d5c42354486.

jenstroeger:
> > In that context, it would also make sense to discuss [`.EXPORT_ALL_VARIABLES:`](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html).
> 
> Not sure what would be the purpose?

See my comment [above](https://github.com/jenstroeger/python-package-template/pull/329#discussion_r989564127) — let’s discuss this separately?

behnazh (approved):

Issue #330, opened 2022-09-21
Skip Pull Request Action on Draft PRs


jenstroeger:
The [pull-request.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/pull-request.yaml) Action may not need to trigger if the PR is in “draft” mode — the [event payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request) has a boolean `"draft"` that we should be able to check. A draft PR by definition is incomplete and may or may not pass checks and tests…

jenstroeger:
@behnazh I think we can close this one now? Two related issues are merged.

jenstroeger:
It would make sense, methinks, to expand the PR Action to check that a PR does not modify certain files. These “protected” files could be stored in the repository in a dedicated [dot-file](https://askubuntu.com/questions/94780/what-are-dot-files), e.g. `.donottouch`.

The intention is to prevent people from modifying these files in downstream repositories, and enforce modifying these files in the parent template repository. Protected files would be e.g. the `Makefile` or `.pre-commit-config.yaml`.

(Related issue #370.)

PR #336, opened 2022-10-04
fix: don’t build anew if a PR was simply edited

jenstroeger:
See also issue #330: there’s no need to run the [build.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/build.yaml) workflow if a PR was merely edited (i.e. no code was changed).


jenstroeger:
Ok, run [3182767662](https://github.com/jenstroeger/python-package-template/actions/runs/3182767662) (6m 36s) was the initial run when we opened this PR which built the artifacts. Then we changed the PR title, thus triggering run [3182834503](https://github.com/jenstroeger/python-package-template/actions/runs/3182834503) (20s). Next, let’s change the title back.

behnazh:
> Hm. I wonder if this is a good idea thinking
> 
> Suppose we create a PR and the [pull-request.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/pull-request.yaml) workflow fails because [build.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/build.yaml) fails; next, we make a comment on that same PR and now the workflow succeeds because it wouldn’t build again…

How about creating two separate workflows for checking PR title and PR code changes to avoid the issue?

jenstroeger:
> How about creating two separate workflows for checking PR title and PR code changes to avoid the issue?

Commit https://github.com/jenstroeger/python-package-template/pull/336/commits/78edf05e524d8b0ddb4b2728555d1d68f6f175d3.

.github/workflows/pr-change-set.yaml
```diff
@@ -0,0 +1,21 @@
+# This workflow checks and tests the package code, and it builds all package
+# artifacts whenever there were changes to a pull request.
+
+name: 'Pull Request: change set'
```
behnazh, line 4: I think `Pull Request` in the name is not necessary because the trigger should appear in the workflow run.

How about "Check change set"?

.github/workflows/pr-conventional-commits.yaml
```diff
@@ -3,7 +3,7 @@
 # tool to check the title of the PR and all commit messages of the branch
 # which triggers this Action.
 
-name: Pull Request
+name: 'Pull Request: conventional commits'
```
behnazh, line 6: Same comment regarding the name.

.github/workflows/pr-change-set.yaml
jenstroeger: On the website the action run shows the `name:` of the Action:

actions

with

https://github.com/jenstroeger/python-package-template/blob/e76f2cdc40a566205c6dfedc07b004f5617261db/.github/workflows/scorecards-analysis.yaml#L3

> How about "Check change set"?

I worry that the run log shows only that, and we don’t have the context that the check runs on a PR?

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/336/commits/bd8d88b09ac1f71ab8d4c78c0a8c7335e3536f94.

.github/workflows/pr-conventional-commits.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/336/commits/bd8d88b09ac1f71ab8d4c78c0a8c7335e3536f94.

.github/workflows/pr-change-set.yaml
```diff
@@ -0,0 +1,21 @@
+# This workflow checks and tests the package code, and it builds all package
+# artifacts whenever there were changes to a pull request.
+
+name: Check change set
+on:
+  pull_request:
+    branches:
+    - main
+    - staging
```
behnazh, line 9: As discussed shall we change the target branches to `*` to make sure the checks run on all pull requests?

jenstroeger: This PR or a separate one?

behnazh (approved):

PR #341, opened 2022-10-11
refactor(ci): allow release when provenance generation fails

behnazh:
This PR decouples the release and provenance generation and allows the release to proceed even if provenance generation fails.


behnazh:
Oops, selected the wrong branch.

PR #342, opened 2022-10-11
refactor(ci): allow release when provenance generation fails

behnazh:
This PR decouples the release and provenance generation and allows the release to proceed even if provenance generation fails.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -165,15 +165,17 @@ jobs:
   # https://github.com/slsa-framework/slsa-github-generator/issues/942
   provenance:
     needs: [build, release]
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.1
     with:
       base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
-      compile-generator: true
+      compile-generator: true # Build the generator from source.
+      # Set private-repository to true for private repositories. Note that the repository name is
+      # uploaded as part of the transparency log entry on the public Rekor instance (rekor.sigstore.dev).
+      private-repository: false
     permissions:
       actions: read # To read the workflow path.
       id-token: write # To sign the provenance.
       contents: write # To add assets to a release.
-
 
   # Publish the SLSA provenance as the GitHub release asset.
   publish_provenance:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,23 +75,13 @@ jobs:
     permissions:
       contents: read
 
-  # Generate the build provenance. The generator should be referenced with a semantic version.
-  # The build will fail if we reference it using the commit SHA.
-  provenance:
-    needs: build
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
-    with:
-      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
-    permissions:
-      actions: read # To read the workflow path.
-      id-token: write # To sign the provenance.
-      contents: write # To add assets to a release.
-
   # Create a new Release on Github from the verified build artifacts, and optionally
   # publish the artifacts to a PyPI server.
   release:
-    needs: [build, provenance]
+    needs: [build]
     name: Release
+    outputs:
+      release-tag: ${{ steps.upload-assets.outputs.release-tag }}
     runs-on: ubuntu-latest
     permissions:
       contents: write # To publish release notes.
@@ -106,11 +96,6 @@ jobs:
       uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
       with:
         fetch-depth: 0
-
-    - name: Download provenance
-      uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # v3.0.0
-      with:
-        name: ${{ needs.provenance.outputs.attestation-name }}
 
     - name: Download artifact
       uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # v3.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -166,3 +151,50 @@ jobs:
...
+  # The build will fail if we reference it using the commit SHA. To avoid using a pre-built
+  # provenance generator which depends on an external service Rekor (https://github.com/sigstore/rekor)
+  # we build this generator from source for now. For more information see this discussion:
+  # https://github.com/slsa-framework/slsa-github-generator/issues/942
+  provenance:
+    needs: [build, release]
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    with:
+      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
+      compile-generator: true
```
jenstroeger, line 165: We’ll keep building the generator from source for now?

.github/workflows/release.yaml
```diff
@@ -166,3 +151,50 @@ jobs:
...
+        fetch-depth: 0
+
+    - name: Download provenance
+      uses: actions/download-artifact@fb598a63ae348fa914e94cd0ff38f362e927b741 # v3.0.0
+      with:
+        name: ${{ needs.provenance.outputs.attestation-name }}
+
+    - name: Upload provenance
+      run: |
+        gh release upload `git describe --tags --abbrev=0` ${{ needs.provenance.outputs.attestation-name }}
```
jenstroeger, line 198: I wonder: if we output the result of `git describe --tags --abbrev=0` from the `release` job and read it here in the `publish_provenance` job, then we wouldn’t need to run the `checkout` action?

Also, it should be safe to write the attestation name into the shell command—we/SLSA folks own that name so there’s no injection issue here?

jenstroeger (commented):
Two questions, bu w00t 🥳

.github/workflows/release.yaml
behnazh: > We’ll keep building the generator from source for now?

Yes.

jenstroeger: Agreed, let’s prune the dependencies 🤓

.github/workflows/release.yaml
behnazh: > I wonder: if we output the result of `git describe --tags --abbrev=0` from the `release` job and read it here in the `publish_provenance` job, then we wouldn’t need to run the `checkout` action?

I think it's a good idea to pass the release tag between jobs, but unfortunately `gh` requires a git repo to work, so we can't avoid checking out the repo.

behnazh: > Also, it should be safe to write the attestation name into the shell command—we/SLSA folks own that name so there’s no injection issue here?

I'm assuming we can trust the attestation name...

behnazh: See commit [3b48cdb](https://github.com/jenstroeger/python-package-template/pull/342/commits/3b48cdb88b7aca55b89729b2c7bdde3bee8364e9).

jenstroeger (approved):
One thought below, but looks good 👍🏼 

To be sure, even if the `provenance` job fails, we’ve still created a Release on Github with all its artifacts. The whole workflow can still fail, though, if the `provenance` job or the `publish_provenance` job fails?

behnazh:
> To be sure, even if the `provenance` job fails, we’ve still created a Release on Github with all its artifacts. The whole workflow can still fail, though, if the `provenance` job or the `publish_provenance` job fails?

That's correct. The version will be bumped, the tag and release will be created, and all the assets except for provenance will be uploaded when provenance generation fails. But the run itself will be shown as failed. I think that's good though to be aware that the provenance is missing.

.github/workflows/release.yaml
```diff
@@ -133,18 +118,24 @@ jobs:
...
     # Create the release including the artifacts and the SLSA L3 provenance.
     - name: Upload assets
+      id: upload-assets
       env:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
-      run: TAG=`git describe --tags --abbrev=0` && gh release create $TAG dist/* ${{ needs.provenance.outputs.attestation-name }} --title $TAG --notes-file RELEASE_NOTES.md
+      run: |
+        TAG=`git describe --tags --abbrev=0`
+        gh release create $TAG dist/* --title $TAG --notes-file RELEASE_NOTES.md
+        echo "release-tag=$TAG" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
behnazh: > This `run` is a single bash script, methinks. So, should the `TAG` name be [exported](https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-export)?

No need to export the env I think?

.github/workflows/release.yaml
```diff
@@ -166,3 +157,47 @@ jobs:
...
     #     TWINE_REPOSITORY_URL=<REPOSITORY_URL>
+
+  # Generate the build provenance. The generator should be referenced with a semantic version.
+  # The build will fail if we reference it using the commit SHA. To avoid using a pre-built
+  # provenance generator which depends on an external service Rekor (https://github.com/sigstore/rekor)
+  # we build this generator from source for now. For more information see this discussion:
+  # https://github.com/slsa-framework/slsa-github-generator/issues/942
+  provenance:
+    needs: [build, release]
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.1
```
jenstroeger, line 168: See also https://github.com/slsa-framework/slsa-github-generator/issues/942#issuecomment-1286374139 👍🏼

.github/workflows/release.yaml
```diff
@@ -133,18 +118,24 @@ jobs:
...
       run: cz changelog --dry-run $(cz version --project) > RELEASE_NOTES.md
 
     # Create the release including the artifacts and the SLSA L3 provenance.
     - name: Upload assets
+      id: upload-assets
       env:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
-      run: TAG=`git describe --tags --abbrev=0` && gh release create $TAG dist/* ${{ needs.provenance.outputs.attestation-name }} --title $TAG --notes-file RELEASE_NOTES.md
+      run: |
+        TAG=`git describe --tags --abbrev=0`
```
jenstroeger, line 136: I ran [actionlint](https://github.com/rhysd/actionlint) and
```
.github/workflows/release.yaml:135:7: shellcheck reported issue in this script: SC2006:style:1:5: Use $(...) notation instead of legacy backticks `...` [shellcheck]
    |
135 |       run: |
    |       ^~~~
.github/workflows/release.yaml:135:7: shellcheck reported issue in this script: SC2086:info:2:19: Double quote to prevent globbing and word splitting [shellcheck]
    |
135 |       run: |
    |       ^~~~
.github/workflows/release.yaml:135:7: shellcheck reported issue in this script: SC2086:info:2:39: Double quote to prevent globbing and word splitting [shellcheck]
    |
135 |       run: |
    |       ^~~~
.github/workflows/release.yaml:135:7: shellcheck reported issue in this script: SC2086:info:3:28: Double quote to prevent globbing and word splitting [shellcheck]
    |
135 |       run: |
    |       ^~~~
```
There are more issues across the actions, so perhaps we should address them all in one separate PR?

.github/workflows/release.yaml
jenstroeger: Also, I think we should add
```yaml
private-repository: false
```
for documentation purposes.

.github/workflows/release.yaml
jenstroeger: Ok I can rebase #348 after this one is merged.

jenstroeger (approved):

PR #345, opened 2022-10-18
fix: exit Makefile gracefully if an inactive venv was found

jenstroeger:
It might happen that a user runs
```
(project) ~/some/path > deactivate
~/some/path > make venv
```
in which case we want to exit gracefully instead of installing a new venv over the existing one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ venv:
 	  echo "Found an activated Python virtual environment, exiting" && exit 1; \
 	fi
 	if [ -d .venv/ ]; then \
-	  echo "Found an inactive Python virtual environment, please activate" && exit 1; \
+	  echo "Found an inactive Python virtual environment, please activate or nuke it" && exit 1; \
 	fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -70,6 +70,9 @@ venv:
 	if [ ! -z "${VIRTUAL_ENV}" ]; then \
 	  echo "Found an activated Python virtual environment, exiting" && exit 1; \
 	fi
+	if [ -d .venv/ ]; then \
+	  echo "Found an inactive Python virtual environment, please activate" && exit 1; \
```
behnazh, line 74: It might be good to ask to "Remove it otherwise."?

jenstroeger: You mean as part of the `echo`?

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/345/commits/c183f13566e3cbd3de2e58b9f859e1ebc42c15f8.

behnazh (approved):

PR #354, opened 2022-10-25
fix: update pytest to drop dependency on vulnerable py package

jenstroeger:
See also: https://github.com/pytest-dev/py/issues/287#issuecomment-1290407715

This PR supersedes https://github.com/jenstroeger/python-package-template/pull/353


jenstroeger:
> I think we should take out `make audit` from `build.yaml` and dedicate a separate workflow for SCA analysis to prevent breaking builds in production.

@behnazh agreed, and we’ll have a draft PR for such a change & discussion soon.

PR #358, opened 2022-10-27
fix(ci): use GITHUB_OUTPUT instead of deprecated set-output

behnazh:


.github/workflows/build.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
         BUILD_EPOCH_PATH=$(find dist/ -type f -name "*-build-epoch.txt")
         DIGEST=$(sha256sum $TARBALL_PATH $WHEEL_PATH $REQUIREMENTS_PATH $SBOM_PATH $HTML_DOCS_PATH $BUILD_EPOCH_PATH | base64 -w0)
         echo "Digest of artifacts is $DIGEST."
-        echo "::set-output name=artifacts-sha256::$DIGEST"
+        echo "artifacts-sha256=$DIGEST" >> $GITHUB_OUTPUT
```
jenstroeger, line 116: Like in PR https://github.com/jenstroeger/python-package-template/pull/348 there should be double-quotes around `$GITHUB_OUTPUT`.

jenstroeger (approved):

PR #372, opened 2022-11-03
feat: Comment Audit result on PR (only on errors)

eduardo-getpassport:
# What's on this PR

- Changes on `pull-request.yaml`:
     - Added step to download the vulnerabilities artifact from `build` step and post a comment with the output.
- Changes on `build.yaml`:
     - Added `continue-on-error: true` in `Audit installed packages` step . In case of an audit error, the next step will run
     - Added the id `audit-packages` in `Audit installed packages` step. This is to reference this step in the later one.
     - Changed the `run` command to export the output.
     - Upload the output artifact

What `Comment vulnerabilities on PR` step does:

Takes the file generated in the previous step in case of an error and posts the content inside the PR.

Example:

Screen Shot 2022-11-03 at 13 21 04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v2.4.2 (2022-10-29)
+
+### Fix
+
+- **ci**: as of v2 Scorecard requires extra permissions (#366)
+
 ## v2.4.1 (2022-10-28)
 
 ### Fix
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,7 @@ update_changelog_on_bump = true
 version_files = [
     "src/package/__init__.py:__version__",
 ]
-version = "2.4.1"
+version = "2.4.2"
 
 
 # https://github.com/pytest-dev/pytest-cov
--- a/src/package/__init__.py
+++ b/src/package/__init__.py
@@ -6,4 +6,4 @@ This package is an empty cookiecutter package example. Enjoy!
 # The version of this package. There's no comprehensive, official list of other
 # magic constants, so we stick with this one only for now. See also this conversation:
 # https://stackoverflow.com/questions/38344848/is-there-a-comprehensive-table-of-pythons-magic-constants
-__version__ = "2.4.1"
+__version__ = "2.4.2"


.github/workflows/build.yaml
```diff
@@ -62,7 +63,15 @@ jobs:
       run: make dist
     # Audit all currently installed packages for security vulnerabilities.
     - name: Audit installed packages
-      run: make audit
+      id: audit-packages
+      continue-on-error: true
+      run: make audit > raw-vulnerabilities.txt && tail -n +5 raw-vulnerabilities.txt > vulnerabilities.txt # Saving the command output and tailing from line 5 to avoid having the command on the PR comment
+    - name: Comment vulnerabilities on PR
+      uses: marocchino/sticky-pull-request-comment@v2
```
jenstroeger, line 70: Can we replace this external action with a call to [`gh issue comment`](https://cli.github.com/manual/gh_issue_comment)?

@eduardo-getpassport I ask because we try to limit our dependencies as much as possible to keep the [software supply chain](https://en.wikipedia.org/wiki/Software_supply_chain) simple.

.github/workflows/build.yaml
```diff
@@ -62,7 +63,15 @@ jobs:
       run: make dist
     # Audit all currently installed packages for security vulnerabilities.
     - name: Audit installed packages
-      run: make audit
+      id: audit-packages
+      continue-on-error: true
+      run: make audit > raw-vulnerabilities.txt && tail -n +5 raw-vulnerabilities.txt > vulnerabilities.txt # Saving the command output and tailing from line 5 to avoid having the command on the PR comment
```
jenstroeger, line 68: Have you tried `make -s` here?

jenstroeger (commented):
This is great, thanks @eduardo-getpassport! A few questions below.

.github/workflows/build.yaml
```diff
@@ -16,6 +16,7 @@ on:
         value: ${{ jobs.build.outputs.artifacts-sha256 }}
 permissions:
   contents: read
+  pull-requests: write
```
behnazh, line 19: We really want to avoid any `write` permissions in `build.yaml` for supply chain security. The reason is that it can build untrusted code from PRs. One thing we could do is to store the results of `make audit` as a job artifact and create a PR comment directly from `pull-request.yaml` in a job that has `write` permission.

.github/workflows/build.yaml
behnazh: Agreed to use `gh` instead of a 3rd party Action. Also, as mentioned in my previous comment, this can be moved to `pull-request.yaml`.

.github/workflows/pull-request.yaml
```diff
@@ -16,6 +16,7 @@ on:
     - synchronize
 permissions:
   contents: read
+  pull-requests: write
```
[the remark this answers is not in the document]
behnazh: Please see my answer in the other comment.

behnazh (changes requested):

.github/workflows/build.yaml
eduardo-getpassport: I replaced this with:

`gh pr comment ${{ github.event.number }} --body-file vulnerabilities.txt`

and it works perfectly! thanks for the suggestion!

(pushing the changes after fixing other things)

.github/workflows/build.yaml
eduardo-getpassport: I didn't realize that! thanks for the suggestion! work's perfectly with the flag 😃

.github/workflows/build.yaml
eduardo-getpassport: I'm working with this step! after doing the testing ill push all the fixes! thanks for the feedback and the explanation 😃

.github/workflows/pull-request.yaml
behnazh, line 19: @eduardo-getpassport At the top we should keep the permissions minimal and just grant the `write` permission to the `comment-pr` job.

.github/workflows/pull-request.yaml
```diff
@@ -63,3 +64,20 @@ jobs:
...
     # See  https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
     # for the security recommendations.
+
+  comment-pr:
+    name: Comment PR with Audit results
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download artifact from Build
+      uses: actions/download-artifact@v3
```
behnazh, line 74: Can you please pin this workflow with the hash digest of the latest like everywhere else in the workflows?

behnazh (changes requested):

PR #384, opened 2022-11-15
feat: Comment Audit result on PR (only on errors)

eduardo-getpassport:
# What's on this PR

- Changes on `pr-change-set.yaml`:
     - Added step to download the vulnerabilities artifact from `build` step and post a comment with the output.
- Changes on `build.yaml`:
     - Added `continue-on-error: true` in `Audit installed packages` step . In case of an audit error, the next step will run
     - Added the id `audit-packages` in `Audit installed packages` step. This is to reference this step in the later one.
     - Changed the `run` command to export the output.
     - Upload the output artifact
- Updated .gitignore with vulnerabilities file to avoid issues.

What `Comment vulnerabilities on PR` step does:

Takes the file generated in the previous step in case of an error and posts the content inside the PR.

Example:

Screen Shot 2022-11-03 at 13 21 04


.github/workflows/pr-conventional-commits.yaml
```diff
@@ -52,4 +52,4 @@ jobs:
       env:
         PR_BASE_REF: ${{ github.event.pull_request.base.ref }}
         PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
-        PR_HEAD_REPO_CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
+        PR_HEAD_REPO_CLONE_URL: ${{ github.event.pull_request.head.repo.clone_url }}
```
eduardo-getpassport, line 55: Don't really know what happened here but the file was no modified.

jenstroeger: Does your editor have last-line linefeed settings?

jenstroeger: Ah: https://github.com/jenstroeger/python-package-template/actions/runs/3474640735/jobs/5808048530#step:10:71

Did the git hooks not run locally on commit and push?

.github/workflows/build.yaml
```diff
@@ -74,7 +74,20 @@ jobs:
...
     # Audit all currently installed packages for security vulnerabilities.
     - name: Audit installed packages
-      run: make audit
+      id: audit-packages
+      continue-on-error: true
+      run: make --silent audit > vulnerabilities.txt
+
+    # Upload the vulnerabilities file output.
+    - name: Upload Artifact
+      id: upload-audit-artifact
```
jenstroeger, line 83: This `id` isn’t referenced so, for simplicity’s sake, we can remove it. Same with the `id: comment-pr` below.

.github/workflows/build.yaml
```diff
@@ -74,7 +74,20 @@ jobs:
...
+      uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb #v3.1.1
+      with:
+        name: vulnerabilities.txt
+        path: .
+        if-no-files-found: error
+        retention-days: 1
+      if: steps.audit-packages.outputs.exit_code == 1
```
jenstroeger, line 90: I wonder if `steps.audit-packages.conclusion == 'failure'` is more reliable because any non-zero exit code is considered a failure (and we check only for `1`<sup>*</sup>)?

—————
<sup>*</sup> At first glance, `1` seems to be the only exit code that [`pip-audit` returns](https://github.com/pypa/pip-audit/blob/main/pip_audit/_cli.py).

.github/workflows/pr-change-set.yaml
```diff
@@ -18,3 +18,23 @@ jobs:
...
+    needs: build
+    steps:
+
+    - name: Download artifact from Build
+      uses: actions/download-artifact@9782bd6a9848b53b110e712e20e42d89988822b7 #v3
+      if: steps.audit-packages.outputs.exit_code == 1
+      with:
+        name: vulnerabilities.txt
+
+    - name: comment PR
```
jenstroeger, line 35: Title case, for consistency: `Comment on PR`

jenstroeger (dismissed):
Looks good to me, thank you @eduardo-getpassport! I have only three minor notes below that you can ignore if you’d like 😉

jenstroeger:
@behnazh I’m actually surprised that referencing workflow steps works across _different_ workflows ([src](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/pr-change-set.yaml#L31)) 🤔 I wasn’t able to find that documented, so hopefully that handy feature isn’t a bug.

For example, in [`pr-change-set.yaml`](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/pr-change-set.yaml#L31) we have:
```yaml
if: steps.audit-packages.outputs.exit_code == 1
```
which references the `audit-packages` step in the reusable workflow [`build.yaml`](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/build.yaml#L75-L79)

behnazh (dismissed):
Thanks, looks great! But the pipelines are failing. I think the problem is storing the result file at the root directory, which `flit` is not tracking and therefore it complains. You can either store it at `dist/` or `/tmp`?

behnazh:
> @behnazh I’m actually surprised that referencing workflow steps works across _different_ workflows ([src](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/pr-change-set.yaml#L31)) thinking I wasn’t able to find that documented, so hopefully that handy feature isn’t a bug.

Good point. I had missed it. Actually I doubt it works and will probably fail at runtime.

@eduardo-getpassport have you tested these workflows?

eduardo-getpassport:
> > @behnazh I’m actually surprised that referencing workflow steps works across _different_ workflows ([src](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/pr-change-set.yaml#L31)) thinking I wasn’t able to find that documented, so hopefully that handy feature isn’t a bug.
> 
> Good point. I had missed it. Actually I doubt it works and will probably fail at runtime.
> 
> @eduardo-getpassport have you tested these workflows?

Yes! I just tested again.

I commented on the `if: steps.audit-packages.outputs.exit_code == 1` from [build]([build.yaml](https://github.com/eduardo-getpassport/python-package-template/blob/486ac05808c10c70acd5477c56c1e01a262b6728/.github/workflows/build.yaml#L75-L79)) and let the two others in the PR file and work as expected (didn't skip the post artifact on the build and skipped in the PR file).

After that I commenter the PR ones and works. Since we are using the `needs: build`, the step can take the ID from another workflow.

eduardo-getpassport:
> Thanks, looks great! But the pipelines are failing. I think the problem is storing the result file at the root directory, which `flit` is not tracking and therefore it complains. You can either store it at `dist/` or `/tmp`?

Fixed on separate branch, pushing the changes rn

.github/workflows/build.yaml
eduardo-getpassport: Removed!

.github/workflows/pr-change-set.yaml
eduardo-getpassport: Adjusted the name!

behnazh:
> Yes! I just tested again.
> 
> I commented on the `if: steps.audit-packages.outputs.exit_code == 1` from build and let the two others in the PR file and work as expected (didn't skip the post artifact on the build and skipped in the PR file).
> 
> After that I commenter the PR ones and works. I think since we are using the `needs: build`, the step can take the ID from another workflow.

@eduardo-getpassport I'm trying to debug this. The `outputs.exit_code ` is not defined anywhere in `build.yaml`  and I can't find any documentation for it. Why are you using this variable? When I run  `echo ${{ steps.audit-packages.outputs.exit_code }}` it's empty for me.

jenstroeger:
With PR https://github.com/jenstroeger/python-package-template/pull/551 in mind I wonder if it would make sense to post an informative message on a PR if package auditing is disabled?

jenstroeger:
@eduardo-getpassport do you mind if I rebase and push to this PR?

PR #387, opened 2022-11-17
fix: determine package version even in deactivated venv

jenstroeger:
Thank you @behnazh for noticing the regression with [v2.4.1](https://github.com/jenstroeger/python-package-template/releases/tag/v2.4.1), introduced by PR https://github.com/jenstroeger/python-package-template/pull/346. The problem is (still and again 🤦🏻‍♂️) that we don’t set up a venv in a Github runner and therefore we shouldn’t rely on the `VIRTUAL_ENV` environment variable.

This change expands the Python command so that the command _always_ returns a package version string (which may be `"unknown"`) instead of failing the `import package`.

I think it’s safe to expect `"unknown"` for `make venv` (or its Action equivalent)

https://github.com/jenstroeger/python-package-template/blob/7bf474b781efa1fb0e69dcf31f968675695f10bc/.github/workflows/build.yaml#L70-L71

and `make setup`

https://github.com/jenstroeger/python-package-template/blob/7bf474b781efa1fb0e69dcf31f968675695f10bc/.github/workflows/build.yaml#L72-L73

because the package hasn’t been installed yet, and therefore its version can’t be determined.

However, after `make setup` the package’s version can be determined safely—with or without virtual environment. For example, the artifacts attached to workflow run [3488321082](https://github.com/jenstroeger/python-package-template/actions/runs/3488321082) indeed contain the correct package version number again:

pkg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,16 +3,9 @@
 # https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
 SHELL := bash
 
-# Set the package's name and version for use throughout the Makefile. If the
-# version can not be determined then error out.
+# Set the package's name and version for use throughout the Makefile.
 PACKAGE_NAME := package
-ifneq ($(origin VIRTUAL_ENV),undefined)
-  PACKAGE_VERSION := $(shell python -c 'import $(PACKAGE_NAME); print($(PACKAGE_NAME).__version__)')
-else ifneq ($(wildcard .venv/upgraded-on),)
-  PACKAGE_VERSION := $(shell .venv/bin/python -c 'import $(PACKAGE_NAME); print($(PACKAGE_NAME).__version__)')
-else
-  $(error Unable to determine package version)
-endif
+PACKAGE_VERSION := $(shell python -c $$'try: import $(PACKAGE_NAME); print($(PACKAGE_NAME).__version__);\nexcept: print("unknown");')
 
 # This variable contains the first goal that matches any of the listed goals
 # here, else it contains an empty string. The net effect is to filter out
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,15 @@
 # https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
 SHELL := bash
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,10 +5,12 @@ SHELL := bash
 
 # Set the package's name and version for use throughout the Makefile.
 PACKAGE_NAME := package
-ifeq ($(origin VIRTUAL_ENV),undefined)
-  PACKAGE_VERSION := unknown
-else
+ifneq ($(origin VIRTUAL_ENV),undefined)
   PACKAGE_VERSION := $(shell python -c 'import $(PACKAGE_NAME); print($(PACKAGE_NAME).__version__)')
+else ifneq ($(wildcard .venv/upgraded-on),)
+  PACKAGE_VERSION := $(shell .venv/bin/python -c 'import $(PACKAGE_NAME); print($(PACKAGE_NAME).__version__)')
```
jenstroeger, line 11: I guess if we use `.venv/bin/python` to determine the package version then we should use the [`$(wildcard)`](https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html) function to check for `python`, instead of the `.venv/upgraded-on` timestamp?

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/387/commits/30641a024e2791a03080ce8aa0b3775f87178477.

jenstroeger:
@behnazh after our conversation today we could add after these two lines

https://github.com/jenstroeger/python-package-template/blob/66d2e0dd1eb63d4e81935cbb2009dc9b7965bfe1/Makefile#L6-L8

the following
```make
ifeq ($(PACKAGE_VERSION),unknown)
  $(warning Unable to determine package version, proceeding anyway)
endif
```
That way, a user shouldn’t be surprised anymore if `"unknown"` sneaks into file names.

behnazh (approved):

behnazh:
> We could add more checks to exclude these two goals, or issue a warning instead of error & exit.

No, I think the code as it is captures the issues and we don't need more warnings.

behnazh (approved):

PR #398, opened 2022-11-26
fix(ci): make Release Notification a reusable workflow to avoid artifact race

### LINKED ISSUE #355: Race condition when releasing a package.
When our `release.yaml` creates a Release ot Github

https://github.com/jenstroeger/python-package-template/blob/f41b0e6a46061081370253b9ac17a1b62b5c85dc/.github/workflows/release.yaml#L122-L126

we enter a race condition where the [Github Release event](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#releaseevent) fires _before_ this current workflow finishes. Which can become an issue because other actions that trigger on that event (like our [`release-notification.yaml`](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/release-notifications.yaml)) may want to access the artifacts attached to this workflow… and that fails because this workflow is still running.

jenstroeger:
If triggered by the [`release[released]`](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#releaseevent) event then the [Release Notification workflow](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/release-notifications.yaml) ([code](https://github.com/jenstroeger/python-package-template/blob/7bf474b781efa1fb0e69dcf31f968675695f10bc/.github/workflows/release.yaml#L137)) starts running before the [Release workflow](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/release.yaml) finishes. That can lead to a race condition where the release artifacts of the Release workflow are not yet available to workflows triggered by the `release[released]` event.

Using a reusable workflow addresses that race condition and enables us to pass artifacts more easily.

Closes #355

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -209,7 +209,9 @@ jobs:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
 
   # Send out release notifications after the Release was published on GitHub.
+  # Uncomment the `if` to disable sending release notifications.
   notifications:
+    # if: ${{ false }}
     needs: [release]
     name: Send Release notifications
     uses: ./.github/workflows/_release-notifications.yaml
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -167,7 +167,6 @@ jobs:
   # https://github.com/slsa-framework/slsa-github-generator/issues/942
   provenance:
     needs: [build, release]
-    name: Generate provenance
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.2
     with:
       base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -82,6 +82,7 @@ jobs:
     name: Release
     outputs:
       release-tag: ${{ steps.upload-assets.outputs.release-tag }}
+      release-url: ${{ steps.upload-assets.outputs.release-url }}
     runs-on: ubuntu-latest
     permissions:
       contents: write # To publish release notes.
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -205,3 +207,17 @@ jobs:
...
+    name: Send Release notifications
+    uses: ./.github/workflows/_release-notifications.yaml
+    permissions:
+      contents: read
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      release_tag: ${{ needs.release.outputs.release-tag }}
+      release_url: ${{ needs.release.outputs.release-url }}
+    secrets:
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
```
jenstroeger, line 223: The original Release Notification was disabled:

https://github.com/jenstroeger/python-package-template/blob/7bf474b781efa1fb0e69dcf31f968675695f10bc/.github/workflows/release-notifications.yaml#L15-L17

This one is not because without the required [Slack Webhook URL](https://api.slack.com/messaging/webhooks) secret it won’t run anyway. (Can’t find docs anymore that said that Actions without sufficient secrets don’t run.)

behnazh: > This one is not because without the required [Slack Webhook URL](https://api.slack.com/messaging/webhooks) secret it won’t run anyway. (Can’t find docs anymore that said that Actions without sufficient secrets don’t run.)

I think it's better to be explicit and add the condition back. Also it would be good to add a note to README.md that if secrets are not set, the workflow might not work at all.

behnazh (approved):

.github/workflows/release.yaml
jenstroeger: Makes sense, commit https://github.com/jenstroeger/python-package-template/pull/398/commits/e70c60770d2b7bb4951286f03c0a3d258bf9c788.

Issue #389, opened 2022-11-18
Consider adding a .gitattributes file.


jenstroeger:
Based on [this conversation](https://github.com/jenstroeger/python-package-template/pull/388#issuecomment-1319592490) with @thbeu we should consider adding a [`.gitattributes`](https://www.git-scm.com/docs/gitattributes) file for better control over e.g. file endings and large file management.

jenstroeger:
@thbeu would you like to add that file and open a PR?

thbeu:
Sure. Can do.

jenstroeger:
Commit https://github.com/jenstroeger/python-package-template/commit/c56f95291db98afc57d8539f7b8b03734a9c50af.

PR #407, opened 2022-11-30
feat: Add .gitattributes file

thbeu:
The .gitattributes file follows the best practices from https://rehansaeed.com/gitattributes-best-practices/

This is for #389.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+# Force batch scripts to always use CRLF line endings so that if a repo is accessed
+# in Windows via a file share from Linux, the scripts will work.
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
+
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,11 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+# Force batch scripts to always use CRLF line endings so that if a repo is accessed
+# in Windows via a file share from Linux, the scripts will work.
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
```
jenstroeger, line 7: I wonder if we should add `.ps1` ([PowerShell](https://learn.microsoft.com/en-us/powershell/)) here?

jenstroeger (commented):
Thanks @thbeu, looks good!

.gitattributes
thbeu: right, will do

jenstroeger (approved):

behnazh (approved):

PR #420, opened 2022-12-13
chore: use long options in Makefile recipes if possible

jenstroeger:


jenstroeger:
One day, when macOS catches up to the 1990s, we’ll reopen this PR 🤦🏻‍♂️

PR #428, opened 2022-12-21
fix(ci): don’t fail bump job if there are no commits to bump

### LINKED ISSUE #422: Merging to main without bump fails CI.
For example:
```
Run cz bump --changelog --yes
bump: release 0.10.5 → 0.10.5

Automatically generated by Commitizen.

tag to create: v0.10.5
[NO_COMMITS_TO_BUMP]
The commits found are not eligible to be bumped

Error: Process completed with exit code 21
```
We probably shouldn’t fail CI here. However, looks like `cz` doesn’t support this on [bump](https://commitizen-tools.github.io/commitizen/bump/) 🤔

jenstroeger:
Closes #422

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,7 @@ jobs:
         git config --list --global # For debug purposes.
 
     - name: Create changelog and bump
-      run: cz bump --changelog --yes
+      run: cz --no-raise 21 bump --changelog --yes
 
     - name: Push the release
       run: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -60,6 +60,9 @@ jobs:
         git config --global user.email "$USER_EMAIL"
         git config --list --global # For debug purposes.
 
+    # In some cases a user may merge commits that don't cause a version bump, which causes commitizen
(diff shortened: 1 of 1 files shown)


eduardok-getpassport (approved):

.github/workflows/release.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
         git config --list --global # For debug purposes.
 
     - name: Create changelog and bump
-      run: cz bump --changelog --yes
+      run: cz --no-raise 21 bump --changelog --yes
```
behnazh, line 64: @jenstroeger Can you please add a comment that:

- Commitizen should not fail when the changelog doesn't have a commit that bumps a version 
- "NoneIncrementExit == 21"
- reference to the Commitizen's [documentation](https://commitizen-tools.github.io/commitizen/bump/#avoid-raising-errors)

behnazh (changes requested):

.github/workflows/release.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/428/commits/7ec5434c1ec847eb87711a203b680a77532ff91d.

behnazh (approved):

PR #437, opened 2023-01-03
chore(test): add Hypothesis command-line switches and configuration profiles

jenstroeger:


.github/workflows/_build.yaml
```diff
@@ -81,6 +81,8 @@ jobs:
     # those targets first and only builds the package if they succeed.
     - name: Build the package
       run: make dist
+      env:
```
behnazh, line 84: Can you please add a comment and clarify this is needed to prevent randomized testing in GitHub?

behnazh (changes requested):

.github/workflows/_build.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/437/commits/a5f1b7b503f22d80d717f3d144815d2f10b81646.

behnazh (approved):

PR #444, opened 2023-01-10
feat: add a workflow to create sync PRs

behnazh:
This PR adds a GitHub Actions workflow, which checks for new releases in this template repo daily and creates a pull request in the downstream repo.

### DIFF
--- a/.github/workflows/sync-with-upstream.yaml
+++ b/.github/workflows/sync-with-upstream.yaml
@@ -26,6 +26,9 @@ jobs:
     - name: Check out template repository
       uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       with:
+        # If you decide to change the upstream template repository to a private one, uncomment
+        # the following argument to pass the required token to be able to check it out.
+        # token: ${{ secrets.<TEMPLATE_REPO_TOKEN> }}
         repository: jenstroeger/python-package-template
         fetch-depth: 0
         path: template
--- a/.github/workflows/sync-with-upstream.yaml
+++ b/.github/workflows/sync-with-upstream.yaml
@@ -1,0 +1,78 @@
+# Create a PR to sync with the upstream template repo.
+# The template repo is https://github.com/jenstroeger/python-package-template.
+
+name: Sync with template repository
+on:
+  schedule:
+  - cron: 11 11 * * *
+
+permissions:
+  contents: read
+env:
+  # This is the username and email for the user who creates a branch and commits
+  # the changes. In an organisation that should be a dedicated devops account.
+  USER_NAME: jenstroeger
+  USER_EMAIL: jenstroeger@users.noreply.github.com
+
+jobs:
+  sync:
+    if: github.repository != 'jenstroeger/python-package-template'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+
+    - name: Check out template repository
+      uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+      with:
+        repository: jenstroeger/python-package-template
+        fetch-depth: 0
+        path: template
+
+    - name: Check out current repository
+      uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+      with:
+        token: ${{ secrets.REPO_ACCESS_TOKEN }}
+        fetch-depth: 0
+        ref: main
+        path: repo
+
+    - name: Sync with template
+      env:
+        GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
+      run: |
+        LATEST_VERSION=$(cd template && git describe --tags --abbrev=0)
+        CURRENT_VERSION=$(test -f repo/.github/workflows/.template_version \
+          && cat repo/.github/workflows/.template_version || echo "v0.0.0")
+        echo "Latest version is $LATEST_VERSION and current version is $CURRENT_VERSION."
+
+        if [ $CURRENT_VERSION == $LATEST_VERSION ]; then
+          echo "Unable to find a new version, exiting..."
+        else
+
+          # Copy the files from the template to the current repo.
+          rsync --recursive --verbose --verbose --exclude tests/ --exclude src/ \
+            --exclude .git/ --exclude .github/workflows/.template_version template/ repo/
+
+          # Check if the branch exists in the current repo.
+          cd repo/
+          if [ $(git rev-parse --verify origin/sync-$LATEST_VERSION 2>/dev/null) ]; then
+            echo "Branch sync-$LATEST_VERSION already exists. Exiting..."
+            exit 0
+          fi
+
+          # Create a branch, commit, and push the changeset.
+          git checkout -b sync-$LATEST_VERSION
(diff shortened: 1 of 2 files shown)


brad-getpassport:
Initial thoughts is that it looks good, I will try and integrate this into our private stuff and test it out and see how things work.

.github/workflows/sync-with-upstream.yaml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+
+    - name: Check out template repository
+      uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+      with:
+        repository: jenstroeger/python-package-template
```
brad-getpassport, line 29: Need the `token: ${{ secrets.... }}` here to make this work, you can't checkout without the token

brad-getpassport (changes requested):

.github/workflows/sync-with-upstream.yaml
behnazh: > Need the `token: ${{ secrets.... }}` here to make this work, you can't checkout without the token

Have you tried this workflow with this [template repo](https://github.com/jenstroeger/python-package-template)? The token should not be needed for checking out this template repo because it's public.

jenstroeger: I think the `token:` is required if the upstream template repository is private. Maybe we should leave a comment about that in this workflow?

behnazh: > I think the `token:` is required if the upstream template repository is private. Maybe we should leave a comment about that in this workflow?

Commit [3c4a557](https://github.com/jenstroeger/python-package-template/pull/444/commits/3c4a5570df074d3d4e99efaa0aee1a22f7b87628).

.github/workflows/sync-with-upstream.yaml
```diff
@@ -0,0 +1,81 @@
...
+          && cat repo/.github/workflows/.template_version || echo "v0.0.0")
+        echo "Latest version is $LATEST_VERSION and current version is $CURRENT_VERSION."
+
+        if [ $CURRENT_VERSION == $LATEST_VERSION ]; then
+          echo "Unable to find a new version, exiting..."
+        else
+
+          # Copy the files from the template to the current repo.
+          rsync --recursive --verbose --verbose --exclude tests/ --exclude src/ \
+            --exclude .git/ --exclude .github/workflows/.template_version template/ repo/
```
jenstroeger, line 59: I’ve used [`rsync`](https://linux.die.net/man/1/rsync) in the past but unfortunately I’m not deeply familiar with _all_ of its features.

My main concern about using `rsync` is that it clobbers the local files, and that then results in a larger change set and possible merge conflicts. I’ve been scratching my head about that for a while now, and wanted to share an alternative to the `rsync` approach:

- Check out the downstream repository as usual, and not into a subfolder;
- Add this upstream template repository as a remote ([docs](https://git-scm.com/docs/git-remote#Documentation/git-remote.txt-emaddem)):
    ```bash
    git remote add upstream https://github.com/jenstroeger/python-package-template.git
    ```
- Merge from the upstream repository and ask `git` to refer upstream changes ([docs](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-ort)):
    ```bash
    git merge --squash --allow-unrelated-histories --strategy ort --strategy-option theirs <REF>  # Where <REF> can be `main` or some other git ref.
    ```
    By using the [`theirs`](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-theirs) option for the default merge strategy `ort` we tell `git` to prefer the upstream changes.

Judging from a few local tests, this allowed me to merge the upstream changes into an independent private repository without merge conflicts; however, the resulting change set lost a few local changes, e.g. the package name of my repository whose merge conflict was resolved using the upstream template’s package name. But maybe that’s ok, because we need to review _and adjust_ the PR anyway 🤔

.github/workflows/sync-with-upstream.yaml
```diff
@@ -0,0 +1,81 @@
...
+          git add .
+          git config --global user.name $USER_NAME
+          git config --global user.email $USER_EMAIL
+          git config --list --global # For debug purposes.
+          git commit -m "chore: sync with template"
+          git push --set-upstream origin sync-$LATEST_VERSION
+
+          # Create the pull request.
+          gh pr create --base main --title "chore: sync with package template $LATEST_VERSION" \
+            --body "This PR was automatically generated." --head sync-$LATEST_VERSION
```
jenstroeger, line 80: Should we add the [`--draft`](https://cli.github.com/manual/gh_pr_create#options) option here?

README.md
```diff
@@ -20,6 +20,7 @@ This repository is intended to be a base template, a cookiecutter for a new Pyth
 [Git hooks](#git-hooks)  
 [Testing](#testing)  
 [Generating documentation](#generating-documentation)  
+[Synchronizing with this template repo](#Synchronizing-with-this-template-repo)
```
jenstroeger, line 23: I think (not sure without checking) that the generated heading id is all lower-case: `#synchronizing-with-this-template-repo`

jenstroeger (commented):
I apologize for the delay, things have been busy these last few weeks. @brad-getpassport has tested the workflow already and it works 🥳

Nonetheless, I wanted to raise a discussion on how we bring over files from an upstream template repository below…

.github/workflows/sync-with-upstream.yaml
jenstroeger: Might be worth digging around [here](https://stackoverflow.com/questions/6557467/can-git-ignore-a-specific-line) or [here](https://stackoverflow.com/questions/16244969/how-to-tell-git-to-ignore-individual-lines-i-e-gitignore-for-specific-lines-of/16244970).

Also, regarding files/path we want to _exclude_ from the automatic merge (e.g. [here](https://stackoverflow.com/questions/6294355/git-how-to-exclude-files-from-merging)): that doesn’t really seem to be doable with `git` but perhaps a simple `git restore` after the above merge would do?

behnazh: These are interesting alternatives to explore, but I prefer to keep it as it is for now and incrementally improve.

jenstroeger: I created issue https://github.com/jenstroeger/python-package-template/issues/472 from the above comment.

.github/workflows/sync-with-upstream.yaml
behnazh: No because `--draft` implied that the PR is not ready for review.

.github/workflows/sync-with-upstream.yaml
jenstroeger: I also assigned that ticket to you, @behnazh.

.github/workflows/sync-with-upstream.yaml
jenstroeger: Hmph.

README.md
behnazh: Commit [4108c96](https://github.com/jenstroeger/python-package-template/pull/444/commits/4108c96fdffcb4ac39a0bba5abaaaa3ca45c6b27).

jenstroeger (approved):

PR #445, opened 2023-01-15
chore!: drop support for Python 3.9, add support for Python 3.11

jenstroeger:


behnazh (approved):

.github/workflows/_build.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
         # It is recommended to pin a Runner version specifically:
         # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python: ['3.9', '3.10']
+        python: ['3.10', '3.11']
```
behnazh, line 47: What's the point of supporting two Python versions only? Wouldn't it make sense to use packages like [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) to manage the Python version and OS?

jenstroeger: I’m not familiar with that package, but it looks interesting. Let’s move your suggestion to a separate issue?

behnazh (approved):

.github/workflows/_build.yaml
behnazh: > I’m not familiar with that package, but it looks interesting. Let’s move your suggestion to a separate issue?

Sounds good, thanks.

jenstroeger: Issue https://github.com/jenstroeger/python-package-template/issues/451.

jenstroeger:
@behnazh I’m unsure about the breaking change `chore!:` here 🤔 What do you think?

behnazh:
> @behnazh I’m unsure about the breaking change `chore!:` here thinking What do you think?

It's a tricky one because this may or may not be a breaking change depending on the downstream repositories. I would slightly prefer to make it a breaking change to communicate the potential side effects?

jenstroeger:
> It's a tricky one because this may or may not be a breaking change depending on the downstream repositories. I would slightly prefer to make it a breaking change to communicate the potential side effects?

Ok then let’s keep the PR title as-is 👍🏼

PR #455, opened 2023-01-30
fix(ci): update isort to latest fix because of PyCQA/isort/issues/2077

jenstroeger:


behnazh (changes requested):
Can you please update all the `pre-commit` hooks using `pre-commit autoupdate` instead? Also, we should add that to `make upgrade`.

jenstroeger:
> Can you please update all the `pre-commit` hooks using `pre-commit autoupdate` instead? Also, we should add that to `make upgrade`.

Would you mind moving a general pre-commit update into a separate PR? This one is really to fix the build break that Poetry introduced, and which now breaks all Dependabot PRs… so I’d like to merge this ASAP.

behnazh (approved):

PR #460, opened 2023-01-31
chore(deps-dev): update pre-commit tools

jenstroeger:


jenstroeger:
@behnazh I’m tempted to roll back Black to a 22.x.x release because of a disagreeable change that shipped with 23.x.x. See also https://github.com/psf/black/issues/902#issuecomment-1425285264 and issue https://github.com/psf/black/issues/3551.

behnazh (approved):

PR #465, opened 2023-02-08
fix: call explicitly Python3 in Makefile

jenstroeger:
Turns out that on new(er) Macs there is no `python` but a `python3`. I suspect that’s to avoid collisions with user-installed `python2` versions (v2.7 the current, last maintained one).

Leaving this PR as a Draft for discussion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL := bash
 
 # Set the package's name and version for use throughout the Makefile.
 PACKAGE_NAME := package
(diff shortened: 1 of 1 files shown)


behnazh:
> Turns out that on new(er) Macs there is no `python` but a `python3`. I suspect that’s to avoid collisions with user-installed `python2` versions (v2.7 the current, last maintained one).
> 
> Leaving this PR as a Draft for discussion.

Not sure if this should be left to developers to resolve on their machines!

jenstroeger:
Fairnuff. Let me close the PR then…

PR #479, opened 2023-02-22
fix: make Black, Flake8, Sphinx play nice together by tweaking a few Flake8 settings

jenstroeger:
There’s much discussion on how to make Black and Flake8 play nice with each other. There’s also much discussion on how well Flake8 implements [PEP8](https://peps.python.org/pep-0008/).

First, we’ve been using the [`ignore`](https://flake8.pycqa.org/en/6.0.0/user/options.html#cmdoption-flake8-ignore) option

https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.flake8#L8

which defaults to `E121,E123,E126,E226,E24,E704,W503,W504`. By providing our own we essentially re-_enabled_ these other options. Then there is [`extend-ignore`](https://flake8.pycqa.org/en/6.0.0/user/options.html#cmdoption-flake8-extend-ignore) which _extends_ to default list, rather than replacing it.

Second, considering the [Using _Black_ with other tools](https://github.com/psf/black/blob/main/docs/guides/using_black_with_other_tools.md#flake8) chapter, it proposes using
```ini
[flake8]
extend-ignore = E203
```
rather than the `ignore` option.

Perhaps we should consider following that pattern and keep Flake8’s defaults enabled? We already use only the [`extend-select`](https://flake8.pycqa.org/en/6.0.0/user/options.html#cmdoption-flake8-extend-select) option which extends the [`select`](https://flake8.pycqa.org/en/6.0.0/user/options.html#cmdoption-flake8-select) defaults `E,F,W,C90`.

<hr>

Then there is the particular question of handling the Flake8 warnings `E203: whitespace before ‘,’, ‘;’, or ‘:’` and `W503: line break before binary operator`. Both seem to disagree with PEP8 and both are triggered by Black; see also [Why are Flake8’s E203 and W503 violated?](https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated) and issue https://github.com/PyCQA/pycodestyle/issues/373.

This change disabled both warnings, and it enables `W504: line break after binary operator` which is [recommended](https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated) by Black and PEP8 compliant.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,11 +5,13 @@
 # https://github.com/psf/black/blob/main/docs/guides/using_black_with_other_tools.md#flake8
 [flake8]
 
-# Enable Bugbear's extended opinionated checks.
-# https://github.com/PyCQA/flake8-bugbear#how-to-enable-opinionated-warnings
+# Enable a few additional checks.
 #
-# Enable W504: line break after binary operator (Black compliant)
+# https://github.com/PyCQA/flake8-bugbear#how-to-enable-opinionated-warnings
+# B9: Bugbear's extended opinionated checks
+#
 # https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
+# W504: line break after binary operator (Black compliant)
 extend-select = B9, W504
 
 # Disable several warnings that don't play nice with PEP8 or Black,
@@ -32,7 +34,7 @@ per-file-ignores =
 max-line-length = 120
 show-source = true
 
-# Ensure that Flake8 warnings are silenced correctly.
+# Ensure that Flake8 warnings are silenced correctly:
 # https://github.com/plinss/flake8-noqa#options
 noqa-require-code = true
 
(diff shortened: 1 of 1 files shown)


jenstroeger:
Commit https://github.com/jenstroeger/python-package-template/pull/479/commits/09c27e2284ccd68acdc101d8f6179794411899a9 adds a few Sphinx related settings, see [docs](https://github.com/peterjc/flake8-rst-docstrings#configuration).

jenstroeger:
The [`flake8-docstrings`](https://github.com/pycqa/flake8-docstrings#flake8-docstrings) plugin has a `--docstring-convention` option, so should we add
```ini
docstring-convention = google
```
to our `.flake8` configuration?

behnazh (approved):

behnazh:
> The [`flake8-docstrings`](https://github.com/pycqa/flake8-docstrings#flake8-docstrings) plugin has a `--docstring-convention` option, so should we add
> 
> ```ini
> docstring-convention = google
> ```
> 
> to our `.flake8` configuration?

Sure. How about `numpy`?

jenstroeger:
> Sure. How about `numpy`?

Commit https://github.com/jenstroeger/python-package-template/pull/479/commits/9544e24735701f0fd93ed05ee8ea4b5f38f07963.

behnazh (approved):

PR #481, opened 2023-02-25
chore: disable pylint’s Similarities checker because of its noise

jenstroeger:
I wanted to disable `pylint`’s [Similarities](https://pylint.pycqa.org/en/latest/user_guide/checkers/features.html#similarities-checker) because it’s been quite noisy with false positives in other projects where it’s always disabled too. Might as well disable it upstream here in our template?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,7 @@ fail-under = 10.0
 disable = [
     "fixme",
     "line-too-long",  # Replaced by Flake8 Bugbear B950 check.
+    "similarities",
     "too-few-public-methods",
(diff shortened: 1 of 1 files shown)


behnazh:
> I wanted to disable `pylint`’s [Similarities](https://pylint.pycqa.org/en/latest/user_guide/checkers/features.html#similarities-checker) because it’s been quite noisy with false positives in other projects where it’s always disabled too. Might as well disable it upstream here in our template?

It might be useful to find copy pasted code. Can we keep it?

jenstroeger:
Closed as per request by @behnazh

PR #487, opened 2023-02-26
fix(ci): ensure correct epoch for packaged files to enable reproducible builds

### LINKED ISSUE #484: Discuss accuracy of the epoch artifact.
### Problem

Attached to every [release](https://github.com/jenstroeger/python-package-template/releases) is a `build-epoch.txt` file containing the timestamp when the artifacts for a particular release were built, e.g. [package-2.8.0-build-epoch.txt](https://github.com/jenstroeger/python-package-template/releases/download/v2.8.0/package-2.8.0-build-epoch.txt).

We use this `build-epoch.txt` file to pass a build epoch timestamp via the [`SOURCE_DATE_EPOCH` environment variable to the `flit` build backend](https://flit.pypa.io/en/stable/reproducible.html), in order to implement [SLSA verified reproducible builds](https://slsa.dev/spec/v0.1/faq#q-what-about-reproducible-builds).

Running `make` it gets the epoch timestamp when it first evaluates the Makefile:

https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/Makefile#L48-L53

Then `make` runs all checks and tests, and only after all checks and tests were successful does `make` build the distribution artifacts:

https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/Makefile#L172-L186

As the last of the `dist` goals, the epoch timestamp is saved to the `build-epoch.txt` file. However, there are two issues here:

- There can be a significant time difference between the captured epoch timestamp and when the distribution artifacts are actually being built; and
- All build artifacts have their own build timestamps that’s almost guaranteed to be different from the captured epoch timestamp.

### Question

@behnazh I was wondering about the accuracy requirements of that build epoch.

### Proposed solution

One possible solution would be to use the artifact files’ last modified dates and write those to the `build-epoch.txt` file:
```make
dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt:
	echo "Makefile $(SOURCE_DATE_EPOCH)" > dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
	echo "$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl $$(stat -c %Y dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl)" >> dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
	echo "$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz $$(stat -c %Y dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz)" >> dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
	echo "$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip $$(stat -c %Y dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip)" >> dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
	echo "$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-md.zip $$(stat -c %Y dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-md.zip)" >> dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
```
which results in
```
> cat dist/package-2.8.0-build-epoch.txt 
Makefile 1677410832
package-2.8.0-py3-none-any.whl 1677410859
package-2.8.0.tar.gz 1677410859
package-2.8.0-docs-html.zip 1677410860
package-2.8.0-docs-md.zip 1677410861
```

jenstroeger:
Fixes #484 

We still have a problem which might not really be one: timezones. When we pass the [`SOURCE_DATE_EPOCH`](https://flit.pypa.io/en/stable/reproducible.html) environment variable then its value is an [epoch](https://en.wikipedia.org/wiki/Unix_time) which, by definition, is in UTC. However, `flit` sets the package files’ date to _that_ value without accounting for the timezone of the host. For example, with a timestamp of `1677448329` (26. Feb 2023 9:52:09 PM UTC, or 27 Feb 2023 7:52:09 AM AEST) the packaged files receive the following date:
```
> ll package/
total 16K
-rw-r--r-- 1 savage admin 385 Feb 26 21:52 __init__.py
-rw-r--r-- 1 savage admin 496 Feb 26 21:52 __main__.py
-rw-r--r-- 1 savage admin  80 Feb 26 21:52 py.typed
-rw-r--r-- 1 savage admin 268 Feb 26 21:52 something.py
```
on my host with AEST, where I would have expected the adjusted AEST date which was the actual packaging date 🤔

There’s currently no open issue for `flit` to discuss this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,9 +175,9 @@ test:
 .PHONY: dist
 dist: dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-md.zip dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
 dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl: check test
-	flit build --setup-py --format wheel
+	SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) flit build --setup-py --format wheel
 dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz: check test
-	flit build --setup-py --format sdist
+	SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) flit build --setup-py --format sdist
 dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip: docs-html
 	python -m zipfile -c dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip docs/_build/html/
 dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-md.zip: docs-md


behnazh (approved):

jenstroeger:
> I'm not sure why you are expecting an adjusted date on your local machine. I think the way to test it is to see if the reproduced artifact is is identical to the original artifact, regardless of timestamp?

And it continues…

> There is a caveat, however: wheels (which are zip files) include the modification timestamp from each file. This will probably be different on each computer, because it indicates when your local copy of the file was written, not when it was changed in version control. These timestamps can be overridden by the environment variable [SOURCE_DATE_EPOCH](https://flit.pypa.io/en/stable/cmdline.html#envvar-SOURCE_DATE_EPOCH).
> 
> ```
> SOURCE_DATE_EPOCH=$(date +%s)
> flit publish
> # Record the value of SOURCE_DATE_EPOCH in release notes for reproduction
> ```

Which is why we added the `SOURCE_DATE_EPOCH` environment variable to the Makefile in the first place.

behnazh (approved):

PR #493, opened 2023-02-27
fix(ci): remove invalid check from Wiki workflow

### LINKED ISSUE #485: Wiki workflow does not run.
In spite of extensive testing in another private repo, our reusable [Wiki workflow](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/_wiki-documentation.yaml) does not run from the Release workflow:

https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/release.yaml#L234-L253

It’s not that it fails to run or finish, instead the reusable Wiki workflow is simply being [skipped](https://github.com/jenstroeger/python-package-template/actions/runs/4222519143/jobs/7331211073):

wiki

Might this be actually an issue:

https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/_wiki-documentation.yaml#L34

See also [discussions/48632](https://github.com/orgs/community/discussions/48632). Uhm… 🤔

jenstroeger:
Fixes #485 

Based on @airtower-luna’s suggestion [here](https://github.com/orgs/community/discussions/48632#discussioncomment-5124844). (I’ll test separately in a private repo, too.)

### DIFF
--- a/.github/workflows/_wiki-documentation.yaml
+++ b/.github/workflows/_wiki-documentation.yaml
@@ -31,7 +31,6 @@ permissions:
 jobs:
   publish-wiki:
     name: Publish Github Wiki
-    if: github.repository.has_wiki == true
     runs-on: ubuntu-latest
     steps:
 


behnazh:
Interesting. So the issue was using `github.repository` instead of `github.event.repository` in the if condition :smile:

.github/workflows/_wiki-documentation.yaml
```diff
@@ -31,7 +31,6 @@ permissions:
 jobs:
   publish-wiki:
     name: Publish Github Wiki
-    if: github.repository.has_wiki == true
```
behnazh, line 34: How about:
```suggestion
if: github.event.repository.has_wiki == true
```

If we remove the condition, would the release fail if the repo has no wiki?

jenstroeger: I’ve not tried it but would there be an `event` object (which contains a properly shaped `repository` object) in a reusable workflow?

jenstroeger: Works 👍🏼

jenstroeger:
Duplicate of #505

PR #482, opened 2023-02-25
fix(ci): add docs/source/ folder to the excludes of the Sync Upstream workflow

jenstroeger:
Considering we have the previous version of the template:

https://github.com/jenstroeger/python-package-template/blob/6b337a397b37d47e26f9b2889882e6e996b8aa9d/.github/workflows/sync-with-upstream.yaml#L49-L50

would it make sense to check if a branch for that previous version exists, and close the PR and delete that now older branch? Or, if not, mark the older PR as [a duplicate of the new PR](https://docs.github.com/en/issues/tracking-your-work-with-issues/marking-issues-or-pull-requests-as-a-duplicate)?

The new branch that’s being opened should include all changes from that older branch 🤔

Also, should we add [`--reviewer`](https://cli.github.com/manual/gh_pr_create#options) to the PR creation?

PR #505, opened 2023-03-21
feat: add actionlint hook and fix current errors

behnazh:
This PR adds [actionlint](https://github.com/Mateusz-Grzelinski/actionlint-py) as a pre-commit hook to lint and check GitHub Actions workflows. The check runs using `make check-actionlint`. 

The new `actionlint` hook detects errors that were also addressed in https://github.com/jenstroeger/python-package-template/pull/482 and https://github.com/jenstroeger/python-package-template/pull/493, which had to be fixed here for the workflows to succeed.

### DIFF
--- a/.github/workflows/sync-with-upstream.yaml
+++ b/.github/workflows/sync-with-upstream.yaml
@@ -50,7 +50,7 @@ jobs:
           && cat repo/.github/workflows/.template_version || echo "v0.0.0")
         echo "Latest version is $LATEST_VERSION and current version is $CURRENT_VERSION."
 
-        if [ $CURRENT_VERSION == $LATEST_VERSION ]; then
+        if [ "$CURRENT_VERSION" == "$LATEST_VERSION" ]; then
           echo "Unable to find a new version, exiting..."
         else
 
@@ -59,23 +59,23 @@ jobs:
             --exclude .git/ --exclude .github/workflows/.template_version template/ repo/
 
           # Check if the branch exists in the current repo.
-          cd repo/
-          if [ $(git rev-parse --verify origin/sync-$LATEST_VERSION 2>/dev/null) ]; then
+          cd repo/ || exit 1
+          if [ "$(git rev-parse --verify origin/sync-"""$LATEST_VERSION""" 2>/dev/null)" ]; then
             echo "Branch sync-$LATEST_VERSION already exists. Exiting..."
             exit 0
           fi
 
           # Create a branch, commit, and push the changeset.
-          git checkout -b sync-$LATEST_VERSION
-          echo $LATEST_VERSION > .github/workflows/.template_version
+          git checkout -b sync-"$LATEST_VERSION"
+          echo "$LATEST_VERSION" > .github/workflows/.template_version
           git add .
-          git config --global user.name $USER_NAME
-          git config --global user.email $USER_EMAIL
+          git config --global user.name "$USER_NAME"
+          git config --global user.email "$USER_EMAIL"
           git config --list --global # For debug purposes.
           git commit -m "chore: sync with template"
-          git push --set-upstream origin sync-$LATEST_VERSION
+          git push --set-upstream origin sync-"$LATEST_VERSION"
 
           # Create the pull request.
           gh pr create --base main --title "chore: sync with package template $LATEST_VERSION" \
-            --body "This PR was automatically generated." --head sync-$LATEST_VERSION
+            --body "This PR was automatically generated." --head sync-"$LATEST_VERSION"
         fi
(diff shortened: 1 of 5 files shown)


behnazh:
Ah, I [forgot](https://github.com/jenstroeger/python-package-template/actions/runs/4484604267/jobs/7885285239?pr=505) again that we don't create `venv` in GitHub Actions :facepalm:

jenstroeger:
Duplicates #493, fixes #485

jenstroeger:
What about something like [actionlint-py](https://github.com/Mateusz-Grzelinski/actionlint-py)?

behnazh:
> What about something like [actionlint-py](https://github.com/Mateusz-Grzelinski/actionlint-py) and [shellcheck-py](https://github.com/shellcheck-py/shellcheck-py)?

Thanks. I knew about `shellcheck-py` but didn't know about `actionlint-py`. I switched to `actionlint-py` , see commit [1bf019e](https://github.com/jenstroeger/python-package-template/pull/505/commits/1bf019e79947521e5097280d601622257fb760a1).

I'm not sure if adding `shellcheck-py` makes sense in this repo though because we don't have any bash scripts :thinking:

jenstroeger:
> I'm not sure if adding `shellcheck-py` makes sense in this repo though because we don't have any bash scripts 🤔

I think it does: `actionlint` can use `shellcheck` to check the job scripts, provided `shellcheck` is available ([docs](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-shellcheck-integ)).

behnazh:
> > I'm not sure if adding `shellcheck-py` makes sense in this repo though because we don't have any bash scripts thinking
> 
> I think it does: `actionlint` can use `shellcheck` to check the job scripts, provided `shellcheck` is available ([docs](https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-shellcheck-integ)).

I have already made a note on that in the `README.md` [here](https://github.com/jenstroeger/python-package-template/pull/505/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R146). Not sure if you had seen it?

.github/workflows/sync-with-upstream.yaml
```diff
@@ -59,23 +59,23 @@ jobs:
             --exclude .git/ --exclude .github/workflows/.template_version template/ repo/
 
           # Check if the branch exists in the current repo.
-          cd repo/
-          if [ $(git rev-parse --verify origin/sync-$LATEST_VERSION 2>/dev/null) ]; then
+          cd repo/ || exit 1
+          if [ "$(git rev-parse --verify origin/sync-"""$LATEST_VERSION""" 2>/dev/null)" ]; then
             echo "Branch sync-$LATEST_VERSION already exists. Exiting..."
             exit 0
           fi
```
jenstroeger, line 66: Hm, so in
```bash
cd repo/ || exit 1
```
the script exits without a message and with return code `1`, and in the next line it exits with a message and return code `0`. Perhaps we can keep things a bit more consistent?

.github/workflows/sync-with-upstream.yaml
```diff
@@ -59,23 +59,23 @@ jobs:
...
-          echo $LATEST_VERSION > .github/workflows/.template_version
+          git checkout -b sync-"$LATEST_VERSION"
+          echo "$LATEST_VERSION" > .github/workflows/.template_version
           git add .
-          git config --global user.name $USER_NAME
-          git config --global user.email $USER_EMAIL
+          git config --global user.name "$USER_NAME"
+          git config --global user.email "$USER_EMAIL"
           git config --list --global # For debug purposes.
           git commit -m "chore: sync with template"
```
jenstroeger, line 75: While we’re making changes here, can we change `-m` to `--message` as well, please?

jenstroeger (commented):
Looks good, thanks @behnazh! Two minor comments follow…

.github/workflows/sync-with-upstream.yaml
behnazh: See commit [3dd239a](https://github.com/jenstroeger/python-package-template/pull/505/commits/3dd239a6f2cec3327a94ff788e658632b6a1f5d6).

.github/workflows/sync-with-upstream.yaml
behnazh: See commit [3dd239a](https://github.com/jenstroeger/python-package-template/pull/505/commits/3dd239a6f2cec3327a94ff788e658632b6a1f5d6).

jenstroeger (approved):

PR #512, opened 2023-04-02
feat: add flake8-pyi plugin to check .pyi files (if any)

### LINKED ISSUE #494: Should we consider `.pyi` files at all?
Packages can inline type annotations into the code, or they can use [`.pyi` stub files](https://peps.python.org/pep-0484/#stub-files) to add type annotations for existing `.py` files.

We’ve not considered the latter and perhaps it would make sense to add the [flake8-pyi](https://github.com/PyCQA/flake8-pyi) plugin for those users who do?

jenstroeger:
Closes #494

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,7 +62,7 @@ repos:
     name: Check flake8 issues
     files: ^src/package/|^tests/
     types: [text, python]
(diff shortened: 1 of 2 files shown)


behnazh:
> @behnazh should mention this in the README?

Sure.

behnazh (approved):

Issue #472, opened 2023-02-14
Revisit how we create sync PRs


jenstroeger:
I’ve used [`rsync`](https://linux.die.net/man/1/rsync) in the past but unfortunately I’m not deeply familiar with _all_ of its features.

My main concern about using `rsync` is that it clobbers the local files, and that then results in a larger change set and possible merge conflicts. I’ve been scratching my head about that for a while now, and wanted to share an alternative to the `rsync` approach:

- Check out the downstream repository as usual, and not into a subfolder;
- Add this upstream template repository as a remote ([docs](https://git-scm.com/docs/git-remote#Documentation/git-remote.txt-emaddem)):
    ```bash
    git remote add upstream https://github.com/jenstroeger/python-package-template.git
    ```
- Merge from the upstream repository and ask `git` to prefer upstream changes ([docs](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-ort)):
    ```bash
    git merge --squash --allow-unrelated-histories --no-commit --strategy ort --strategy-option theirs <REF>  # Where <REF> can be `main` or some other git ref.
    ```
    By using the [`theirs`](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-theirs) option for the default [merge strategy `ort`](https://git-scm.com/docs/merge-strategies) we tell `git` to prefer the upstream changes for merge conflicts.

Judging from a few local tests, this allowed me to merge the upstream changes into an independent private repository without merge conflicts; however, the resulting change set lost a few local changes, e.g. the package name of my repository whose merge conflict was resolved using the upstream template’s package name. But maybe that’s ok, because we need to review _and adjust_ the PR anyway 🤔

I do wonder if it’s possible to configure the [merge driver](https://git-scm.com/docs/gitattributes.html#_performing_a_three_way_merge) such that I can change the strategy option of the `ort` strategy for certain hunks: for example, the Makefile should merge using the `theirs` option except for the hunk covering line 7

https://github.com/jenstroeger/python-package-template/blob/4e0a9aa60d0fe31775cef87a9cb8fc7c3299e218/Makefile#L7

where I’d like to either exclude a merge completely or merge with the `ours` option.

It might also be worth digging around [here](https://stackoverflow.com/questions/6557467/can-git-ignore-a-specific-line) or [here](https://stackoverflow.com/questions/16244969/how-to-tell-git-to-ignore-individual-lines-i-e-gitignore-for-specific-lines-of/16244970).

[...]

Issue #509, opened 2023-04-02
Tag a previous Upstream Sync PR as duplicate of a new Sync PR


jenstroeger:
> > would it make sense to check if a branch for that previous version exists, and close the PR and delete that now older branch? Or, if not, mark the older PR as [a duplicate of the new PR](https://docs.github.com/en/issues/tracking-your-work-with-issues/marking-issues-or-pull-requests-as-a-duplicate)?
> >
> > The new branch that’s being opened should include all changes from that older branch🤔

Yes, the new PR should technically supersede the previous one and close it.

_Originally posted by @behnazh in https://github.com/jenstroeger/python-package-template/issues/482#issuecomment-1493439563_

PR #537, opened 2023-05-05
fix(ci): improve automatic template synchronization

jenstroeger:
Fixes https://github.com/jenstroeger/python-package-template/issues/472.

Tagging @eduardocardoso because he wrote most of the original workflow.

@behnazh Do we want to add issue https://github.com/jenstroeger/python-package-template/issues/509 to this change as well?

### DIFF
--- a/.github/workflows/sync-with-upstream.yaml
+++ b/.github/workflows/sync-with-upstream.yaml
@@ -80,11 +80,11 @@ jobs:
             git config --global user.name "$USER_NAME"
             git config --global user.email "$USER_EMAIL"
             git config --list --global # For debug purposes.
-            git commit --message "chore(deps): sync with template $LATEST_VERSION"
+            git commit --message "chore: sync with template $LATEST_VERSION"
             git push --set-upstream origin "$BRANCH_NAME"
 
             # Create the pull request.
-            gh pr create --base staging --title "chore(deps): sync with package template $LATEST_VERSION" \
+            gh pr create --base staging --title "chore: sync with package template $LATEST_VERSION" \
               --body "This PR was generated automatically." --head "$BRANCH_NAME"
 
           fi
--- a/.github/workflows/sync-with-upstream.yaml
+++ b/.github/workflows/sync-with-upstream.yaml
@@ -38,7 +38,7 @@ jobs:
       with:
         token: ${{ secrets.REPO_ACCESS_TOKEN }}
         fetch-depth: 0
-        ref: main
+        ref: staging
         path: repo
 
     - name: Sync with template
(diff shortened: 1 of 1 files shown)


eduardocardoso (approved):

jenstroeger:
Github defines a `ready_for_review` PR activity type ([docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)) which runs a workflow when the PR is ready (i.e. not in Draft mode).[^1]

I wonder if we should add that to our pull-request.yaml workflow [here](https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/pr-change-set.yaml#L9-L12), then open the auto-sync PRs in Draft mode [here](https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/sync-with-upstream.yaml#L78-L80) — to avoid running the workflows, failing them (because they always do if there were merge conflicts), to avoid sending spam emails about failed PRs whenever we update the upstream template.

[^1]: I think we discussed previously that we may not want to run workflows on Draft PRs 🤔

.github/workflows/sync-with-upstream.yaml
```diff
@@ -50,33 +50,41 @@ jobs:
           && cat repo/.github/workflows/.template_version || echo "v0.0.0")
         echo "Latest version is $LATEST_VERSION and current version is $CURRENT_VERSION."
 
-        if [ $CURRENT_VERSION == $LATEST_VERSION ]; then
+        # Check if the template repo was changed/updated.
+        if [ "$CURRENT_VERSION" == "$LATEST_VERSION" ]; then
           echo "Unable to find a new version, exiting..."
-        else
```
behnazh, line 55: Can you please keep the `else` to have an explicit `if ... else ...`  pattern?

jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/537/commits/0f96f6eaa30a9dcab5f3d2e23bf0024160aed280.

.github/workflows/sync-with-upstream.yaml
```diff
@@ -50,34 +50,42 @@ jobs:
...
+            pushd template
+            git diff "$CURRENT_VERSION..$LATEST_VERSION" > diff.patch
+            popd
 
-          # Check if the branch exists in the current repo.
-          if [ "$(git rev-parse --verify origin/sync-"""$LATEST_VERSION""" 2>/dev/null)" ]; then
-            echo "Branch sync-$LATEST_VERSION already exists. Exiting..."
-            exit 0
-          fi
+            # Apply the generated patch to the current repo.
```
behnazh, line 69: Can you please add a comment to explain the purpose of the next few lines?

jenstroeger: You think folks aren’t familiar with the standard [`patch`](https://linux.die.net/man/1/patch) command?

behnazh: Apart from the `patch` command, certain files are excluded, which needs further clarification.

.github/workflows/sync-with-upstream.yaml
```diff
@@ -50,34 +50,42 @@ jobs:
...
+            pushd repo
+            patch --strip 1 --batch --merge --input ../template/diff.patch || true
+            git checkout .github/workflows/.template_version src/ docs/ tests/ "*.md"
```
jenstroeger, line 72: @eduardocardoso @behnazh Instead of creating a “full” patch here

https://github.com/jenstroeger/python-package-template/blob/7a1f03fce75fde6c1a0b7807c8dbbf0bb6e3834e/.github/workflows/sync-with-upstream.yaml#L66

then using `patch` (with likely unwanted changes) and then undoing the unwanted changes using `git checkout` — why not generate a patch with only the necessary changes?

I think something like this might work:
```bash
git diff "$CURRENT_VERSION..$LATEST_VERSION" $(find . .github/workflows/ -maxdepth 1 -type f -name '*' ! -name '*.md' ! -name '.template_version') > diff.patch
```
It would also make explicit to the user which files will be part of the patch, and therefore which files are sync’ed automatically from the upstream template.

.github/workflows/sync-with-upstream.yaml
jenstroeger: @behnazh resolving this conversation because of comment https://github.com/jenstroeger/python-package-template/pull/537#discussion_r1267961971 which, I think, addresses the issue differently and more clearly.

.github/workflows/sync-with-upstream.yaml
eduardocardoso: Makes sense. There's no reason to include those files in the patch

behnazh: Please add `docs/` to the command to include `docs/Makefile`.

jenstroeger: Thanks @behnazh, commit https://github.com/jenstroeger/python-package-template/pull/537/commits/2aed4287fe0b7f6b968a1e8a5022150a82ae752e.

behnazh (approved):

PR #543, opened 2023-05-11
fix(make): improve targets and prerequisites to avoid running a recipe unnecessarily

jenstroeger:


eduardocardoso (dismissed):

Makefile
```diff
@@ -162,14 +169,18 @@ check-mypy:
 	pre-commit run mypy --all-files
 check-actionlint:
 	pre-commit run actionlint --all-files
-check:
+check: .venv/checked-on
+.venv/checked-on: $(PACKAGE_DEPS)
```
behnazh, line 173: Looks like we are introducing back `pre-commit` running on staged files only (instead of all files)?

jenstroeger: `PACKAGE_DEPS` contains all current Python files in the `src/package/` folder, and if any of those changes since `make` last touched `.venv/checked-on` then we run the recipe again. Whether they’re staged or not shouldn’t make a difference.

jenstroeger:
I’m closing this PR because it adds unnecessary complexity and dependencies to the Makefile; I think it “solves” a non-issue.

Issue #542, opened 2023-05-10
Improve scripted building


jenstroeger:
We currently still have small blobs of code in our Action workflows that could move into the Makefile. The goal of this exercise is to be able to use the Makefile locally and be able to build and publish a package release locally. The Action workflows should then simply use these Makefile goals, too.

Here’s a list of new Makefile goals we can consider:

- `make dist-hash` https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/_build.yaml#L107-L122
- `make release` https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/release.yaml#L65-L74
- `make release-publish` https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/release.yaml#L148-L167
- `make release-gh` https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/release.yaml#L134-L146 and https://github.com/jenstroeger/python-package-template/blob/0b40f48f607bf1a69b8f4963dea3105ee0f3bf93/.github/workflows/release.yaml#L213-L216

The [_release-notifications.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/_release-notifications.yaml) and [_wiki-documentation.yaml](https://github.com/jenstroeger/python-package-template/blob/main/.github/workflows/_wiki-documentation.yaml) Actions, in their entirety, could also be hoisted into the Makefile and be made part of the release.

PR #544, opened 2023-05-11
feat(make): add the emergency-release target to locally build and publish a new package release without Github Action workflows

jenstroeger:
Closes https://github.com/jenstroeger/python-package-template/issues/542

However, this new `make` target does not consider — in a way even duplicates! — how the Action workflows proceed. Unless we consider PR https://github.com/jenstroeger/python-package-template/pull/537, trying to lower more workflow jobs into the Makefile would result in running checks and tests too often.

Thought: should we make this target

https://github.com/jenstroeger/python-package-template/blob/f8d7a1872d350ae6805b1cceac21fe80f63292e8/Makefile#L210

dependent on `check test` (or just `dist`) to ensure the package is alright?


behnazh:
While I understand the need the frustration if GitHub goes down, I'm afraid adding this step goes against the SLSA and build integrity principles. I'm not sure if it's a good idea to make manual publishing possible in this repo. But in an emergency case, people can always publish manually if they want to anyway.

jenstroeger:
Maybe something to raise with the SLSA folks to discuss? How they’d recommend dealing with such a case?

I actually thought about adding a section to the README regarding workflow assumptions — one, the required infrastructre needs to be available and functional (e.g. Github/CI, Sigstore, …); and two, the tools used must be functional and trusted (e.g. commitizen).

The other day we had workflows fail half-way-through because of a bug in commitizen, which meant we could bump the git repo correctly but failed to generate all artifacts.

PR #552, opened 2023-05-23
fix(deps): temporarily pin typing-extensions deps, to avoid breaking the sphinxnotes-markdown-builder package

jenstroeger:
See also https://github.com/dgilland/pydash/issues/197


behnazh (approved):

jenstroeger:
@behnazh I think we might be stuck with this dependency: https://github.com/dgilland/pydash/issues/197#issuecomment-1567530083 😞

PR #589, opened 2023-07-28
feat: add pip cache

gabrielpiassetta:
Add pip cache to Setup python step

### DIFF
--- a/.github/workflows/_build.yaml
+++ b/.github/workflows/_build.yaml
@@ -67,6 +67,7 @@ jobs:
       uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4.7.0
       with:
         python-version: ${{ matrix.python }}
+        cache: pip
 
     # Using the Makefile assumes an activated virtual environment, which doesn't exist
     # when running in an Action environment (https://github.com/actions/setup-python/issues/359).


jenstroeger:
Just to be sure: as [per docs](https://github.com/actions/setup-python#caching-packages-dependencies) the
```yaml
cache: 'pip'
```
seems to cache [`pip`’s internal cache](https://pip.pypa.io/en/latest/topics/caching/) between workflows (and therefore between workflow VMs) of the _same_ run — but that’s not stated explicitly. Is that correct?

Furthermore, note that [SLSA Build Level 3](https://slsa.dev/spec/v1.0/levels#build-l3-hardened-builds) requires to

> prevent runs from influencing one another, even within the same project.

The term “run” isn’t clearly defined, but we need to keep in mind that while the public PyPI artifactory is trusted, caches in between _are not_ because modifications/tempering is invisible to us (see also [cache poisoning](https://en.wikipedia.org/wiki/Cache_poisoning)). (And that includes an [AWS Artifactory](https://aws.amazon.com/codeartifact/) or the like!)

Also, looking at the [runtimes of the `Check change set` workflow ](https://github.com/jenstroeger/python-package-template/actions/workflows/pr-change-set.yaml) I get values (random sampling) ranging from 3m 55s, 4m 25s to 5m 31s, 6m 17s and up to 7m 7s, 8m 8s — with this change [clocking in at 7m 45s](https://github.com/jenstroeger/python-package-template/actions/runs/5695241797):

Screenshot 2023-09-02 at 17 24 41

So, do we have a measurable impact resulting from this?

jenstroeger:
@behnazh what do you think of this comment: https://github.com/pre-commit/pre-commit/issues/2847#issuecomment-1514683897

behnazh:
> @behnazh what do you think of this comment: [pre-commit/pre-commit#2847 (comment)](https://github.com/pre-commit/pre-commit/issues/2847#issuecomment-1514683897)

This solution seems to have similar issues for build isolation. We have designed `_build.yaml` reusable workflow in a way to be fully isolated, e.g., by limiting the inputs to the workflow and not using any third-party GitHub Actions (except for the trusted Actions from GitHub itself). Enabling caching breaks the `isolation` security property.

Issue #5, opened 2021-10-11
Consider code complexity tracking, and other tools.


jenstroeger:
Tools like [Wily](https://wily.readthedocs.io/) and [Radon](https://radon.readthedocs.io/) allow to track code complexity metrics, e.g. based on [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) or [Halstaed’s complexity metrics](https://en.wikipedia.org/wiki/Halstead_complexity_measures). It might make sense to track these (and other?) metrics to measure _software maintainability_.

In addition, e.g. [Bandit](https://github.com/PyCQA/bandit) or [dlint](https://github.com/dlint-py/dlint) might be useful additions?

jenstroeger:
See also the [**Complexity** section of Awesome flake8 plugins](https://github.com/DmytroLitvinov/awesome-flake8-extensions#complexity).

> In addition, e.g. [Bandit](https://github.com/PyCQA/bandit) or [dlint](https://github.com/dlint-py/dlint) might be useful additions?

PR #214 adds Bandit as got commit hook.

jenstroeger:
It probably makes sense to consider the [pip-audit](https://github.com/trailofbits/pip-audit) package too (see [discussion](https://discuss.python.org/t/towards-a-pip-audit-subcommand-for-vulnerability-analysis-management/17681)).

jenstroeger:
> It probably makes sense to consider the [pip-audit](https://github.com/trailofbits/pip-audit) package too (see [discussion](https://discuss.python.org/t/towards-a-pip-audit-subcommand-for-vulnerability-analysis-management/17681)).

Played around with it, and the tools looks useful:
```
> pip-audit 
Found 2 known vulnerabilities in 2 packages
Name     Version ID             Fix Versions
-------- ------- -------------- ------------
lxml     4.7.1   PYSEC-2022-230 4.9.1
waitress 2.1.1   PYSEC-2022-205 2.1.2
```
However, I wouldn’t use it as a commit hook because dependecies don’t change that often. Instead, it would probably make sense to add it to our Makefile:

https://github.com/jenstroeger/python-package-template/blob/c7541754294c634cd807068ffaf37bd0279cbb52/Makefile#L80-L86

Call `git-audit` right after the packages have been updated:
```makefile
	python -m pip_audit
```

jenstroeger:
Moving PR https://github.com/jenstroeger/python-package-template/pull/377 over here:

[...]

PR #593, opened 2023-08-02
feat: introduce the Ruff pre-commit hook to replace a bunch of other code formatters and linter hooks

jenstroeger:
We have talked about [`ruff`](https://beta.ruff.rs/) before, so this adds an initial setup to play around with. See also https://github.com/jenstroeger/python-package-template/issues/5#issuecomment-1521568770.

Even though `ruff` is supposed to be a [drop-in replacement for Flake8](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-flake8) it doesn’t currently support all of the Flake8 plugins we use in this template. I guess for now we therefore keep `flake8` around. The current `ruff` configuration is [derived](https://pypi.org/project/flake8-to-ruff/) from the current [.flake8](https://github.com/jenstroeger/python-package-template/blob/main/.flake8) file, but we could consider expanding it.


behnazh:
What are the additional checks that `ruff` offers, which would otherwise be missed by our checks?

copdips:
ruff could replace almost all the python linters for now, only thing that it can not do for the moment as per my experience, is the typing check, so we could combine ruff with mypy or pyright.

jenstroeger:
I agree @copdips, and thank you for the comment!

The [list of rules](https://docs.astral.sh/ruff/rules/) that `ruff` implements (based on a multitude of pylint and flake8 plugins) is huge, and if you want please feel free to contribute settings which at least meet the current settings of this repository, maybe even improve it. The important thing we’d like to avoid by adding `ruff` is to regress because we’ve dropped a few checks…

jenstroeger:
Interestingly, git 2.54.0 improved its hooks ([docs](https://git-scm.com/docs/git-hook#_description)) so perhaps using ruff and with some massaging we can get rid of pre-commit entirely… 🤔

jenstroeger:
@behnazh I updated this PR with the latest version of `ruff`, but see also issue https://github.com/astral-sh/ruff/issues/970. Still need to ~update the README and~ run a few more tests, but perhaps it‘s time to switch… 🤔

PR #632, opened 2023-10-25
feat: add support for Python 3.12

jenstroeger:
@behnazh I’ve not yet dropped support for Python 3.10 because we’re using it in other projects.

Also, I noticed that Dependabot — when checking for package updates — doesn’t consider the Python version? What if a package dropped support for some Python version or has not yet added support for 3.12?


behnazh:
> I’ve not yet dropped support for Python 3.10 because we’re using it in other projects.

No, I'm already using Python 3.11. Let's update to Python 3.12.

>Also, I noticed that Dependabot — when checking for package updates — doesn’t consider the Python version? What if a package dropped support for some Python version or has not yet added support for 3.12?

I think the best we can do is to rely on the tests passing after the update.

jenstroeger:
> No, I'm already using Python 3.11. Let's update to Python 3.12.

This change makes the package default to Python 3.12. Are you suggesting to drop support for Python 3.10 completely?

> I think the best we can do is to rely on the tests passing after the update.

Fairnuff.

Also, I can now reproduce the 3.12 check failures locally and will look into it… Might be that some packages haven’t caught up quite yet 🤔

behnazh:
> This change makes the package default to Python 3.12. Are you suggesting to drop support for Python 3.10 completely?

Yes, I'm fine with that.

behnazh (approved):

jenstroeger:
@behnazh and now there’s [pre-release of 3.13](https://docs.python.org/3.13/whatsnew/3.13.html)… Separate PR?

behnazh:
> @behnazh and now there’s [pre-release of 3.13](https://docs.python.org/3.13/whatsnew/3.13.html)… Separate PR?

Shall we wait for the official release?

PR #637, opened 2023-11-12
feat: run doctest as part of running tests, which collects doctests from both the package’s doc strings and the package documentation

jenstroeger:
So far we’ve had [doctest](https://docs.python.org/3/library/doctest.html) support somewhat configured, but it wasn’t actually used because doctest didn’t run. This PR adds full support for running doctest on both

- Python package docstrings, and
- `.rst` files in the documentation.

Tasks to discuss and (maybe) to add to this PR:

- [x] Should doctests contribute to the overall code coverage? If so, we’d probably have to combine/append the results of both coverage runs ([docs](https://coverage.readthedocs.io/en/latest/index.html)) for a final report; not sure how this is integrated into pytest ([docs](https://github.com/pytest-dev/pytest-cov)).
- [x] We may want to mention doctest support in the README?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ test = [
     "pytest >=7.2.0,<8.0.0",
     "pytest-custom_exit_code ==0.3.0",
     "pytest-cov ==4.1.0",
-    "pytest-doctestplus ==1.0.0",
+    "pytest-doctestplus ==1.1.0",
     "pytest-env ==1.1.1",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -208,6 +208,18 @@ max-line-length = 120
 # https://docs.pytest.org/en/latest/reference/customize.html#configuration-file-formats
 # https://docs.pytest.org/en/latest/reference/reference.html#configuration-options
 # https://docs.pytest.org/en/latest/reference/reference.html#command-line-flags
+#
+# To integrate Hypothesis into pytest and coverage, we use its native plugin:
+# https://hypothesis.readthedocs.io/en/latest/details.html#the-hypothesis-pytest-plugin
+#
+# To discover tests in documentation, we use doctest and the doctest-plus plugin which
+# adds multiple useful options to control tests in documentation. More details at:
+# https://docs.python.org/3/library/doctest.html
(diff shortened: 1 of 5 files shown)


behnazh:
> * Should doctests contribute to the overall code coverage? If so, we’d probably have to combine/append the results of both coverage runs ([docs](https://coverage.readthedocs.io/en/latest/index.html)) for a final report; not sure how this is integrated into pytest ([docs](https://github.com/pytest-dev/pytest-cov)).

Yes, that would  be a good idea but not sure how to do it neither :thinking:

pyproject.toml
```diff
@@ -209,11 +210,14 @@ max-line-length = 120
...
 [tool.pytest.ini_options]
 minversion = "7.0"
-addopts = "-vv --doctest-modules --tb native --hypothesis-show-statistics --hypothesis-explain --hypothesis-verbosity verbose -ra --cov package"  # Consider adding --pdb
+addopts = """-vv -ra --tb native \
+    --hypothesis-show-statistics --hypothesis-explain --hypothesis-verbosity verbose \
+    --doctest-modules --doctest-continue-on-failure --doctest-glob '*.rst' --doctest-plus \
+    --suppress-no-test-exit-code \
+    --cov package \
+"""  # Consider adding --pdb
+# https://docs.python.org/3/library/doctest.html#option-flags
```
behnazh, line 219:
```suggestion
# https://docs.pytest.org/en/latest/reference/reference.html#command-line-flags
# https://docs.python.org/3/library/doctest.html#option-flags
[tool.pytest.ini_options]
minversion = "7.0"
addopts = "-vv --doctest-modules --tb native --hypothesis-show-statistics --hypothesis-explain --hypothesis-verbosity verbose -ra --cov package"  # Consider adding --pdb
addopts = """-vv -ra --tb native \
    --hypothesis-show-statistics --hypothesis-explain --hypothesis-verbosity verbose \
    --doctest-modules --doctest-continue-on-failure --doctest-glob '*.rst' --doctest-plus \
    --suppress-no-test-exit-code \
    --cov package \
"""  # Consider adding --pdb
```

behnazh (approved):

PR #667, opened 2023-12-25
feat(ci): switch from building the SLSA provenance generator to using the pre-built version to improve runtime performance

### LINKED ISSUE #663: Switch to pre-build SLSA provenance generator
Some time ago with PR https://github.com/jenstroeger/python-package-template/pull/342 we switched to building the SLSA provenance generator from source; in particular @behnazh commented [here](https://github.com/jenstroeger/python-package-template/pull/342#discussion_r995151214) on the problem:

> […] in general I would prefer to build from source, but we can revisit when the patched generator is released.

That was for v1.2.0 of the [SLSA provenance generator](https://github.com/slsa-framework/slsa-github-generator/blob/e9626e1e3645ffe2d0e279f5d48d0d63f0f6387f/.github/workflows/generator_generic_slsa3.yml#L150), whose current version is v1.9.0.

Sampling through some workflow runs of the past few months:
```bash
for RUN in $( gh run list --workflow 'Check and Release' --branch main --json databaseId --jq '.[].databaseId' ); do
  read -r FROM TO <<<$( gh run view $RUN --json jobs --jq '.jobs[] | select(.name=="provenance / generator") | "\(.startedAt) \(.completedAt)"'); 
  echo $RUN took $((`date -d "$TO" +%s` - `date -d "$FROM" +%s`)); 
done
```
it looks like the `generator` job takes anywhere between [1m 27s](https://github.com/jenstroeger/python-package-template/actions/runs/6954016492/job/18920204300) and [3m 07s](https://github.com/jenstroeger/python-package-template/actions/runs/5630362713/job/15256321151), in other repos I observed very similar fluctuations often leaning towards the three minutes.

I think we should revisit the decision to build the generator from source and consider using the pre-built binary, i.e. change

https://github.com/jenstroeger/python-package-template/blob/40e438b0adab9e4d90f754811bba7a1284d592d9/.github/workflows/release.yaml#L183

to `false`.

jenstroeger:
Closes #663

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -180,7 +180,7 @@ jobs:
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
     with:
       base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
-      compile-generator: true # Build the generator from source.
+      compile-generator: false # Do not build the provenance generator from source anymore.
       # Set private-repository to true for private repositories. Note that the repository name is
       # uploaded as part of the transparency log entry on the public Rekor instance (rekor.sigstore.dev).
       private-repository: false


jenstroeger:
Data from related production packages: the `release / provenance` step is down

- from 2m 29s to 18s, and
- from 2m 12s to 15s

for two different package releases.

behnazh (approved):

PR #669, opened 2023-12-25
chore(ci): add comment to skip check and test and build for Draft PRs because often times they’re unfinished

jenstroeger:


.github/workflows/_build.yaml
```diff
@@ -39,6 +39,7 @@ env:
 
 jobs:
   build:
+    if: github.event.pull_request.draft == false
```
behnazh, line 42: I have found it very useful to build on draft PRs. So, I'm not sure I agree with this change.

behnazh (changes requested):

behnazh:
I'm not sure if this is a `fix` PR? :thinking:

.github/workflows/_build.yaml
jenstroeger: In larger organizations where multiple people work on the same branch it’s common practice to _intentionally_ push changes that don’t pass checks & tests for safekeeping or to share current changes.

Furthermore, for private repositories where organizations pay for Github minutes, running checks & tests on changes that are known to fail can add up to significant amounts of money.

The purpose of a “draft” PR is to signal work in progress or, as Github [puts it](https://github.blog/2019-02-14-introducing-draft-pull-requests/), “code isn’t in any state to be judged”. Running checks and tests on unfinished work in progress doesn’t seem useful.

jenstroeger:
> I'm not sure if this is a `fix` PR? 🤔

Are you suggesting to turn the `feat(ci)` into a `fix(ci)` here? What problem would it fix, though?

behnazh:
> > I'm not sure if this is a `fix` PR? 🤔
> 
> Are you suggesting to turn the `feat(ci)` into a `fix(ci)` here? What problem would it fix, though?

I meant `feat`, not `fix`.

.github/workflows/_build.yaml
behnazh: I would argue that such organizations should adjust the workflows based on their requirements.

>The purpose of a “draft” PR is to signal work in progress or, as Github [puts it](https://github.blog/2019-02-14-introducing-draft-pull-requests/), “code isn’t in any state to be judged”. Running checks and tests on unfinished work in progress doesn’t seem useful.

I disagree with this. Running CI on drat PRs can be very useful. Generally, we don't need to wait until the end to run CI tests.

jenstroeger: Commented out.

behnazh (approved):

PR #918, opened 2025-04-07
fix(ci): make unique and improve Github workflow job ids to make it easier to reference them from Github settings

jenstroeger:
Refs: a4da94f0f1fd7a5d3b15b9d58a8fe5589159f5f4

### DIFF
--- a/.github/workflows/pr-change-set.yaml
+++ b/.github/workflows/pr-change-set.yaml
@@ -21,7 +21,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  pr-build-package:
     uses: ./.github/workflows/_build.yaml
     permissions:
       contents: read
--- a/.github/workflows/pr-change-set.yaml
+++ b/.github/workflows/pr-change-set.yaml
@@ -21,7 +21,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  pr-build-package:
+  build-package:
     uses: ./.github/workflows/_build.yaml
     permissions:
       contents: read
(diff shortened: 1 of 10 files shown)


.github/workflows/pr-change-set.yaml
```diff
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 jobs:
-  build:
+  build-package:
```
behnazh, line 18:
```suggestion
  check-and-build-package:
```

.github/workflows/release.yaml
```diff
@@ -11,7 +11,7 @@ permissions:
   contents: read
 
 jobs:
-  check:
+  check-package-for-release:
```
behnazh, line 14: We also trigger this `job` on `push` to the `main` branch.
```suggestion
  check-and-build-package:
```

.github/workflows/release.yaml
```diff
@@ -21,8 +21,8 @@ jobs:
 
   # On pushes to the 'release' branch create a new release by bumping the version
   # and generating a change log. That's the new bump commit and associated tag.
-  bump:
-    needs: check
+  bump-package-version:
+    needs: [check-package-for-release]
```
behnazh, line 25:
```suggestion
    needs: [check-and-build-package]
```

behnazh (changes requested):

.github/workflows/release.yaml
jenstroeger: Commit https://github.com/jenstroeger/python-package-template/pull/918/commits/2adc4d51b0cf429e08f8735255b452511930ec6d.

behnazh (dismissed):

PR #922, opened 2025-04-14
feat(ci): group multiple Dependabot changes into single PRs for breaking/non-breaking changes

jenstroeger:
Leaving this in Draft mode while testing and to discuss details, e.g. can we name the groups better, do we want to refine the grouping, whatnot…

See also:

- Official [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--), and
- [Grouped version updates for Dependabot public beta](https://github.blog/changelog/2023-06-29-grouped-version-updates-for-dependabot-public-beta/), and
- [Grouped version updates for Dependabot are generally available](https://github.blog/changelog/2023-08-24-grouped-version-updates-for-dependabot-are-generally-available/).

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -14,6 +14,18 @@ updates:
     include: scope
   open-pull-requests-limit: 13
   target-branch: main
+  # Group updates into pull requests. See also:
+  # https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--
+  groups:
+    pip-breaking-changes:
+      applies-to: version-updates
+      update-types:
+      - major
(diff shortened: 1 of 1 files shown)


jenstroeger:
Hmm, in another repository I’m now seeing this error 

Screenshot 2025-04-28 at 14 25 33

because Dependabot is set up ([docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#target-branch-)) to open PRs against a branch _other_ than the default branch 🤔

jenstroeger:
The _disadvantage_ of this approach is to filter out packages that I don’t wish to update.

PR #933, opened 2025-05-02
perf(test): use Python’s system monitoring facilities to improve performance of test runs

jenstroeger:
Based on the [Optimizing coverage with Python 3.12’s sys.monitoring](https://blog.trailofbits.com/2025/05/01/making-pypis-test-suite-81-faster/#optimizing-coverage-with-python-312s-sysmonitoring) blog section, this change almost doubles the performance of test runs (based on trying this change in three production repos).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ test = [
     "pytest >=7.2.0,<9.0.0",
     "pytest-cases ==3.8.6",
     "pytest-custom_exit_code ==0.3.0",
-    "pytest-cov ==6.1.0",
+    "pytest-cov ==6.1.1",  # Uses: coverage[toml] >=7.5
     "pytest-doctestplus ==1.3.0",
     "pytest-env ==1.1.5",
 ]
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ check:
 # the hook setup itself does not pass files to pytest (see .pre-commit-config.yaml).
 .PHONY: test
 test:
-	pre-commit run pytest --hook-stage push --files tests/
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -267,6 +267,10 @@ env = [
 filterwarnings = [
     "error",
     "always::DeprecationWarning",
+    # This warning is issued in two contexts, and we ignore both:
+    # Python 3.10, 3.11: sys.monitoring isn't available in this version, using default core (no-sysmon)
+    # Python 3.12, 3.13: sys.monitoring can't measure branches in this version, using default core (no-sysmon)
+    "always::coverage.exceptions.CoverageWarning",
```
jenstroeger, line 273: We can probably completely ignore the first warning (for Py 3.10 and 3.11) because that’s expected behavior (see [Warnings Filter](https://docs.python.org/3/library/warnings.html#warning-filter). But we should not see the second one warning… 🤔

.pre-commit-config.yaml
```diff
@@ -165,7 +165,7 @@ repos:
   hooks:
   - id: pytest
     name: Run unit tests
-    entry: pytest -c pyproject.toml --cov-config pyproject.toml src/package/ tests/ docs/
+    entry: env COVERAGE_CORE=sysmon pytest -c pyproject.toml --cov-config pyproject.toml src/package/ tests/ docs/
```
behnazh, line 168: Please add a comment based on the PR description to explain why `COVERAGE_CORE=sysmon` is necessary here, and also mention that this option is only recommended for Python version > 3.12.

behnazh (changes requested):

behnazh (approved):

PR #1144, opened 2026-04-20
fix(ci): when gathering build artifacts for subsequent jobs/steps then constrain finding said artifacts and ignore generated directories

jenstroeger:


jenstroeger:
Duplicated by PR https://github.com/jenstroeger/python-package-template/pull/1154, closing.

PR #1152, opened 2026-04-23
fix(deps): skip over crummy mypy version 1.20.2 which has cache issues across multiple platforms

jenstroeger:
Refs: f9ad85ef0dae9f29381d778dcde65307dfd4c7a6


jenstroeger:
Close because #1148 covered this issue.

PR #1220, opened 2026-07-06
feat(ci): cache package venv and pre-commit venvs to improve performance of the _build.yaml workflow

jenstroeger:


jenstroeger:
See also PR https://github.com/jenstroeger/python-package-template/pull/1265 and make sure caching continues to work!

